PR #2, opened 2018-03-24
add a sample.py script

anthrotype:
I was just playing with the api and tried to rewrite the sample script from harfbuzz upstream repo using uharfbuzz:
https://github.com/harfbuzz/harfbuzz/blob/master/src/sample.py

just scrap it if you think we don't need it here

### DIFF
--- a/examples/sample.py
+++ b/examples/sample.py
@@ -1,0 +1,45 @@
+#!/usr/bin/python3
+
+from __future__ import print_function
+import uharfbuzz as hb
+import sys
+
+
+with open(sys.argv[1], 'rb') as fp:
+    fontdata = fp.read()
+
+text = sys.argv[2]
+
+face = hb.Face.create(fontdata, 0)
+font = hb.Font.create(face)
+upem = face.upem
+del face
+
+font.scale = (upem, upem)
+hb.ot_font_set_funcs(font)
+
+buf = hb.Buffer.create()
+
+def debug_message(msg):
+    print(msg)
+    return True
+
+buf.set_message_func(debug_message)
+
+buf.add_str(text)
+buf.guess_segment_properties()
+
+features = {"kern": True, "liga": True}
+hb.shape(font, buf, features)
+del font
+
+infos = buf.glyph_infos
+positions = buf.glyph_positions
+
+for info, pos in zip(infos, positions):
+    gid = info.codepoint
+    cluster = info.cluster
+    x_advance = pos.x_advance
+    x_offset = pos.x_offset
+    y_offset = pos.y_offset
+    print("gid%d=%d@%d,%d+%d" % (gid, cluster, x_advance, x_offset, y_offset))
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -55,6 +55,7 @@ cdef class GlyphPosition:
 
 cdef class Buffer:
     cdef hb_buffer_t* _hb_buffer
+    cdef object _message_func
 
     def __cinit__(self):
         self._hb_buffer = NULL
@@ -110,7 +111,6 @@ cdef class Buffer:
             position.set(glyph_positions[i])
             positions.append(position)
         return positions
-
 
     @property
     def language(self) -> str:
@@ -178,6 +178,23 @@ cdef class Buffer:
     def guess_segment_properties(self) -> None:
         hb_buffer_guess_segment_properties(self._hb_buffer)
 
+    def set_message_func(self, func: Callable[[str], bool]) -> None:
+        # XXX I'm passing the python callback as user_data and recovering it
+        # from the C-API func below. Is there a better way?
+        # Also, the callback is only taking a message argument, no buffer,
+        # nor font or its own user_data. Not sure whether/how to pass them on
+        hb_buffer_set_message_func(
+            self._hb_buffer, _buffer_message_func, <void*>func, NULL)
+        self._message_func = func
+
+
+cdef hb_bool_t _buffer_message_func(hb_buffer_t *buffer, hb_font_t *font,
+                                    const char *message, void *user_data):
+    cdef object py_func = <object>user_data
+    cdef str py_message = (<bytes>message).decode("utf-8")
+    return bool(py_func(py_message))
+
+
 cdef hb_user_data_key_t k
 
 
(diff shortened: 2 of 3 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -209,13 +226,13 @@ cdef class Face:
         if self._hb_face is not NULL:
             hb_face_destroy(self._hb_face)
 
-    """ use bytes/bytearray, not Blob
     @classmethod
-    def create(self, blob: Blob, index: int):
+    def create(cls, fontdata: bytes, index: int):
```
anthrotype: since it's a python function, here we could probably accept only just bytes, but a union of bytes/bytearray/array.array or anything that exposes the buffer interface really.

anthrotype: I meant _not_ only

anthrotype:
yo @adrientetar

examples/sample.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/python3
+
+from __future__ import print_function
```
adrientetar: we don't need this?

examples/sample.py
```diff
@@ -0,0 +1,45 @@
...
+import uharfbuzz as hb
+import sys
+
+
+with open(sys.argv[1], 'rb') as fp:
```
adrientetar: nit: fp → fontfile?

examples/sample.py
```diff
@@ -0,0 +1,45 @@
...
+    fontdata = fp.read()
+
+text = sys.argv[2]
+
+face = hb.Face.create(fontdata, 0)
+font = hb.Font.create(face)
+upem = face.upem
+del face
```
adrientetar: nit: font keeps a reference to face, so "del face" isn't doing much here. Is it here to somehow highlight that face isn't used after that point?

examples/sample.py
```diff
@@ -0,0 +1,45 @@
...
+infos = buf.glyph_infos
+positions = buf.glyph_positions
+
+for info, pos in zip(infos, positions):
+    gid = info.codepoint
+    cluster = info.cluster
+    x_advance = pos.x_advance
+    x_offset = pos.x_offset
+    y_offset = pos.y_offset
+    print("gid%d=%d@%d,%d+%d" % (gid, cluster, x_advance, x_offset, y_offset))
```
adrientetar: nit: could use an f-string ;)

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -178,6 +178,23 @@ cdef class Buffer:
     def guess_segment_properties(self) -> None:
         hb_buffer_guess_segment_properties(self._hb_buffer)
 
+    def set_message_func(self, func: Callable[[str], bool]) -> None:
+        # XXX I'm passing the python callback as user_data and recovering it
+        # from the C-API func below. Is there a better way?
+        # Also, the callback is only taking a message argument, no buffer,
+        # nor font or its own user_data. Not sure whether/how to pass them on
```
adrientetar: We could probably look into this, but do we at all need implementing "message func" in the first place?

The nice thing about the other stuff I put thus far is I use it so I know what things need to be passed to python or not..

src/uharfbuzz/_harfbuzz.pyx
adrientetar: so you're looking for _memoryview_? also can you keep the name blob (and below hb_blob), thus far I keep the same signature names as harfbuzz. (and bytes is a binary "blob" so.. we can revisit later if needed)

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -451,3 +468,7 @@ def ot_layout_table_get_script_tags(face: Face, tag: str) -> List[str]:
         packed = cstr
         tags.append(packed.decode())
     return tags
+
+
+cpdef void ot_font_set_funcs(Font font):
```
adrientetar: Why cpdef here? I used def for all module level functions.

I don't mind cpdef (is it more efficient?) but we should use it consistently in the file then

src/uharfbuzz/charfbuzz.pxd
```diff
@@ -197,3 +203,6 @@ cdef extern from "hb-ot.h":
         unsigned int start_offset,
         unsigned int* script_count,  # in/out
         hb_tag_t* script_tags)  # out
+
+    # hb-ot-font.h
+    void hb_ot_font_set_funcs(hb_font_t *font)
```
adrientetar: nit: so far I put the stars before the space

examples/sample.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/python3
```
adrientetar: I think we could simply put this example in the readme, it's the first file ppl look at so more convenient.

examples/sample.py
anthrotype: not really, was just a blind copy from harfbuzz own example script

src/uharfbuzz/_harfbuzz.pyx
anthrotype: I don't know if we need to implement it. I was just trying to port the simple debugger that is exemplified in the harfbuzz's original sample.py, but using uharfbuzz.

src/uharfbuzz/_harfbuzz.pyx
anthrotype: it's only more efficient in so far as you also need to call a cpdef function from within cython (in which case it uses its faster cdef version). But if we only call this from python, then def is ok.

src/uharfbuzz/charfbuzz.pxd
anthrotype: I copy-pasted it from the harfbuzz headers, which put the stars after the space, but I don't mind.

examples/sample.py
anthrotype: ok

src/uharfbuzz/_harfbuzz.pyx
adrientetar: maybe we can leave it out for now, I've never needed to use it myself

anthrotype: ok

anthrotype: @adrientetar I removed from the example. Shall I remove the set_message_func method altogether from the pyx source file as well?

setup.py
```diff
@@ -1,12 +1,22 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 from skbuild import setup
+import os
+from io import open
```
adrientetar: nit: can you sort the imports

src/uharfbuzz/_harfbuzz.pyx
adrientetar: yes, and pxd too

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -209,13 +226,13 @@ cdef class Face:
         if self._hb_face is not NULL:
             hb_face_destroy(self._hb_face)
 
-    """ use bytes/bytearray, not Blob
     @classmethod
-    def create(self, blob: Blob, index: int):
+    def create(cls, bytes blob, int index=0):
         cdef Face inst = cls()
-        inst._hb_face = hb_face_create(blob, index)
+        cdef hb_blob_t* cblob = hb_blob_create(
```
adrientetar: nit: name it hb_blob

PR #8, opened 2018-11-13
Add hb_font_set_variations func

m4rc1e:
This pr adds the hb function [hb_font_set_variations](https://harfbuzz.github.io/harfbuzz-hb-font.html#hb-font-set-variations).

I've added a method to the font class so users can set variations using a dict in the following manner. 

```
font.variation({"wght": 300, "slnt": -20})
```

I'm unsure whether to keep the method name singular? the hb function is plural.

I need to implement this so I can generate diff images using freetype-py and this lib. If I need more variable related structs and functions, I'll add them. Also, thanks for making this binding. It's the best one imo.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -287,6 +287,26 @@ cdef class Font:
         x, y = value
         hb_font_set_scale(self._hb_font, x, y)
 
+    def set_variations(self, variations: Dict[str, float]) -> None:
+        cdef unsigned int size
+        cdef hb_variation_t* hb_variations
+        cdef bytes packed
+        cdef hb_variation_t variation
+        size = len(variations)
+        hb_variations = <hb_variation_t*>malloc(size * sizeof(hb_variation_t))
+        if not hb_variations:
+            raise MemoryError()
+
+        try:
+            for i, (name, value) in enumerate(variations.items()):
+                packed = name.encode()
+                variation.tag = hb_tag_from_string(packed, -1)
+                variation.value = value
+                hb_variations[i] = variation
+            hb_font_set_variations(self._hb_font, hb_variations, size)
+        finally:
+            free(hb_variations)
+
 
 cdef hb_position_t _glyph_h_advance_func(hb_font_t* font, void* font_data,
                                          hb_codepoint_t glyph,
--- a/src/uharfbuzz/charfbuzz.pxd
+++ b/src/uharfbuzz/charfbuzz.pxd
@@ -141,6 +141,9 @@ cdef extern from "hb.h":
         hb_codepoint_t glyph,
         char *name, unsigned int size,
         void *user_data)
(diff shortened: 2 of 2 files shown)


anthrotype:
`set_variations` would be a better name, as it matches the one from hb

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -287,6 +287,23 @@ cdef class Font:
...
+
+        size = len(variations)
+        hb_variations = <hb_variation_t*>malloc(size * sizeof(hb_variation_t))
+        for i, (name, value) in enumerate(variations.items()):
+            packed = name.encode()
+            variation.tag = hb_tag_from_string(packed, -1)
+            variation.value = value
+            hb_variations[i] = variation
+        hb_font_set_variations(self._hb_font, hb_variations, size)
+        free(hb_variations)
```
anthrotype: maybe wrap the call to hb_font_set_variations in a `try: ... finally: free(hb_variations)` block.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -287,6 +287,23 @@ cdef class Font:
         x, y = value
         hb_font_set_scale(self._hb_font, x, y)
 
+    def variation(self, variations: Dict[str, float]) -> None:
+        cdef unsigned int size
+        cdef hb_variation_t* hb_variations
+        cdef bytes packed
+        cdef char* cstr
```
anthrotype: this cstr variable seems unused

anthrotype:
Thanks Marc, looks good!

src/uharfbuzz/_harfbuzz.pyx
anthrotype: No, that’s not necessary. Actually, since we don’t expect the c++ function to raise an exception the try block is not needed here either, nevermind.

anthrotype: on the other hand, the python statements in between the allocation of hb_variations and the call to free may raise exceptions, in which case we need to release it.
this is the standard pattern when using malloc/calloc/etc. from cython:
https://cython.readthedocs.io/en/latest/src/tutorial/memory_allocation.html
you'll notice that it also checks that malloc returns a NULL pointer in the unlikely event that you ran out of memory

anthrotype:
I just updated the embedded harfbuzz library to the latest version (2.1.1) in the master branch, you may want to first rebase your PR branch on it to check everything still works.

m4rc1e:
Thanks for the review. I've squashed and repushed the following changes:

- method name changed to `set_variations`
- removed the unnecessary variable
- rebased with the latest hb
- implemented the try/finally malloc stuff.

For the last part, I just copied the example from the documentation. I have a vague idea of what's happening but you may want to look at this again.

anthrotype:
tagged 0.3.0 https://github.com/trufont/uharfbuzz/releases/tag/v0.3.0
wheels coming

PR #11, opened 2018-12-10
Unicode cluster index

### LINKED ISSUE #9: info.cluster is utf-8 index, not unicode?
It appears that the info.cluster index is utf-8-based, not unicode character index, as if using `hb-shape --utf8-clusters`. Is this intentional? Intuitively I would have assumed info.cluster would be a unicode char index.

anthrotype:
The returned info.cluster value after using `Buffer.add_str` should now match the indexes of the input python unicode string.
This works with both narrow and wide python builds (python3 is always wide).
The `add_utf8` method can be useful when reading bytes from a UTF-8 text file.

Note I only tested this locally. We need to add a proper test suite at some point.

Fixes #9

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -1,3 +1,4 @@
+#cython: language_level=3
 from charfbuzz cimport *
 from libc.stdlib cimport free, malloc
 from libc.string cimport const_char
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -2,6 +2,11 @@ from charfbuzz cimport *
 from libc.stdlib cimport free, malloc
 from libc.string cimport const_char
 from typing import Callable, Dict, List, Tuple
+import array
+from cpython cimport array
+
+
+cdef bint PY_NARROW_UNICODE = sizeof(Py_UNICODE) != 4
 
 
 cdef class GlyphInfo:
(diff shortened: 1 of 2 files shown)


anthrotype:
cc @khaledhosny @justvanrossum

justvanrossum:
Thanks! Looks good to me.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1,7 +1,13 @@
+#cython: language_level=3
 from charfbuzz cimport *
 from libc.stdlib cimport free, malloc
 from libc.string cimport const_char
 from typing import Callable, Dict, List, Tuple
+import array
+from cpython cimport array
```
adrientetar: Can you maybe sort the imports

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1,7 +1,13 @@
...
+
+
+cdef bint PY_NARROW_UNICODE = sizeof(Py_UNICODE) != 4
```
adrientetar: Python 3 doesn't need this, do we really need this stuff?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -162,17 +168,46 @@ cdef class Buffer:
         if hb_codepoints is not NULL:
             free(hb_codepoints)
 
+    def add_utf8(self, text: bytes,
```
adrientetar: Why bytes here?

src/uharfbuzz/_harfbuzz.pyx
anthrotype: ok

src/uharfbuzz/_harfbuzz.pyx
anthrotype: it doesn't cost us anything to also support python2 when using cython. We keep writing code using python3 language syntax and semantics and cython does the rest.

src/uharfbuzz/_harfbuzz.pyx
anthrotype: what else should it be?

anthrotype: it can't be str (unicode) because the add_utf8 method is intented to be used with UTF-8 encoded bytes. A python unicode string by definition doesn't have any specific "encoding".

If you want I can use a cython memoryview like `const uint8_t[:]` so that it works with any object that supports the buffer interface, bytes, bytearray, array.array, etc.

justvanrossum:
FWIW, I'm not so sure we need this utf8 method at all. uharfbuzz should provide a Pythonic interface to hb, we don't need to be as flexible as hb itself.

khaledhosny:
The utf8 function does not hurt, it should be useful if for whatever reason one is already working with UTF-8 bytes instead of Python strings. May be such a distinction should be emphasized in documentation (if there is any) to avoid any potential confusion that the utf8 method should be used for non-ASCII text.

anthrotype:
I'd like to keep the `add_utf8` method, not just because it matches the existing `hb_buffer_add_utf8` function from harfbuzz API, but because like Khaled said it can be useful when one already is working with UTF-8 bytes.

I simplified the code further by removing the need to copy the string into an array.array and encode it, by using the CPython API to access the string's internal buffer of Py_UNICODE* and passing that directly to harfbuzz.
https://github.com/trufont/uharfbuzz/commit/ebdf61ee3d0a256cb8808219b32b493641e85115

PR #12, opened 2018-12-11
initialize C structs inside __cinit__; deprecate alternative create() constructors

anthrotype:
Following discussions at https://github.com/trufont/uharfbuzz/issues/10#issuecomment-445481033

Buffer, Face, Font and FontFuncs can now be instantiated using the regular python constructor syntax (e.g. `Buffer()` as opposed to `Buffer.create()`)
The old `create` classmethod constructors still work but are deprecated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,12 @@ matrix:
         - PYTHON=python2
         - PIP_INSTALL_OPTIONS="--upgrade --user"
 
+branches:
+  only:
+    # skip PR branches, only build from master and tagged commits
+    - master
+    - /^v\d+\.\d+.*$/
+
 install:
   - $PYTHON -m pip install $PIP_INSTALL_OPTIONS pip
   - $PYTHON -m pip install $PIP_INSTALL_OPTIONS cibuildwheel==0.10.0
--- a/README.md
+++ b/README.md
@@ -19,14 +19,14 @@ with open(sys.argv[1], 'rb') as fontfile:
 
 text = sys.argv[2]
 
-face = hb.Face.create(fontdata)
-font = hb.Font.create(face)
+face = hb.Face(fontdata)
+font = hb.Font(face)
 upem = face.upem
 
 font.scale = (upem, upem)
 hb.ot_font_set_funcs(font)
 
-buf = hb.Buffer.create()
+buf = hb.Buffer()
 
 buf.add_str(text)
 buf.guess_segment_properties()
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,6 +6,12 @@ environment:
     TWINE_PASSWORD:
       secure: H3o/uI5KfKGTFumKKW68hg==
 
+branches:
+  only:
+    # skip PR branches, only build from master and tagged commits
+    - master
+    - /^v\d+\.\d+.*$/
+
 install:
   - python -m pip install --upgrade pip
   - pip install cibuildwheel==0.10.0
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -63,16 +63,16 @@ cdef class Buffer:
     cdef hb_buffer_t* _hb_buffer
 
     def __cinit__(self):
-        self._hb_buffer = NULL
+        self._hb_buffer = hb_buffer_create()
 
     def __dealloc__(self):
         if self._hb_buffer is not NULL:
             hb_buffer_destroy(self._hb_buffer)
 
+    # DEPRECATED: use the normal constructor
     @classmethod
     def create(cls):
         cdef Buffer inst = cls()
-        inst._hb_buffer = hb_buffer_create()
         return inst
 
     @property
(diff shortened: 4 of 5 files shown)


jenskutilek:
https://github.com/trufont/uharfbuzz/blob/dd4aa47137e93d4a0cfdcce0b7c7bd68a9659b9c/src/uharfbuzz/_harfbuzz.pyx#L260

This is now throwing an error because __cinit__() is called without an argument:

```
  File "_harfbuzz.pyx", line 271, in uharfbuzz._harfbuzz.Face.create_for_tables
  File "_harfbuzz.pyx", line 248, in uharfbuzz._harfbuzz.Face.__cinit__
TypeError: __cinit__() takes at least 1 positional argument (0 given)
```

jenskutilek:
I can't get uharfbuzz to build on my system, so I can't try to fix it myself at the moment…

anthrotype:
Ah i see. The blob parameter in cinit method should be set to None by default, so it can be initialised without arguments (in which case the _hn_face attribute should stay NULL).
I’m away from my laptop at the moment. You can pin to an earlier version for the time being.
Why can’t you build from source? What error do you get?

jenskutilek:
I have `cmake` installed, but pip tries to install it anyway as build dependency, which fails …

```bash
$ pip install . --install-option="-std=libc++"
/Users/jens/trufont_env/lib/python3.7/site-packages/pip/_internal/commands/install.py:207: UserWarning: Disabling all use of wheels due to the use of --build-options / --global-options / --install-options.
  cmdoptions.check_install_build_global(options)
Processing /Users/jens/Code/uharfbuzz
  Installing build dependencies ... error
  Complete output from command /Users/jens/trufont_env/bin/python3 /Users/jens/trufont_env/lib/python3.7/site-packages/pip install --ignore-installed --no-user --prefix /private/var/folders/dp/c9fjc9m107v4pqjyprys509h0000gn/T/pip-build-env-9ttp9_ec/overlay --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- "setuptools >= 36.4" "wheel == 0.31.1" "setuptools_scm >= 2.1" "cython >= 0.28.1" "scikit-build >= 0.7.1" cmake ninja:
  Collecting setuptools>=36.4
    Using cached https://files.pythonhosted.org/packages/3b/17/81a4ab4c94e9c78f98dfaad0208994129213089361dc53a7cd8de177d1e3/setuptools-40.7.1.zip
  Collecting wheel==0.31.1
    Using cached https://files.pythonhosted.org/packages/2a/fb/aefe5d5dbc3f4fe1e815bcdb05cbaab19744d201bbc9b59cfa06ec7fc789/wheel-0.31.1.tar.gz
  Collecting setuptools_scm>=2.1
    Using cached https://files.pythonhosted.org/packages/54/85/514ba3ca2a022bddd68819f187ae826986051d130ec5b972076e4f58a9f3/setuptools_scm-3.2.0.tar.gz
  Collecting cython>=0.28.1
    Using cached https://files.pythonhosted.org/packages/02/24/f73045afb049295b34ac55aaf6ea1592604cda3749632a22e563e66604a3/Cython-0.29.3.tar.gz
  Collecting scikit-build>=0.7.1
    Using cached https://files.pythonhosted.org/packages/30/ee/1d6bc8c865b77d7c74487c63770582b72675c5f3260d687f1504996a5379/scikit-build-0.8.1.tar.gz
  Collecting cmake
    Using cached https://files.pythonhosted.org/packages/72/7b/8d3d83f884283ddfc86ca407d10152d66b84dff5baf5ed4537c127f988cb/cmake-3.13.3.tar.gz
      Complete output from command python setup.py egg_info:
        File "<string>", line 1, in <module>
        File "/private/var/folders/dp/c9fjc9m107v4pqjyprys509h0000gn/T/pip-install-ekstrud_/cmake/setup.py", line 7, in <module>
          from skbuild import setup
      ModuleNotFoundError: No module named 'skbuild'
  
      ----------------------------------------
  Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/dp/c9fjc9m107v4pqjyprys509h0000gn/T/pip-install-ekstrud_/cmake/
  
  ----------------------------------------
Command "/Users/jens/trufont_env/bin/python3 /Users/jens/trufont_env/lib/python3.7/site-packages/pip install --ignore-installed --no-user --prefix /private/var/folders/dp/c9fjc9m107v4pqjyprys509h0000gn/T/pip-build-env-9ttp9_ec/overlay --no-warn-script-location --no-binary :all: --only-binary :none: -i https://pypi.org/simple -- "setuptools >= 36.4" "wheel == 0.31.1" "setuptools_scm >= 2.1" "cython >= 0.28.1" "scikit-build >= 0.7.1" cmake ninja" failed with error code 1 in None
```

anthrotype:
The libc++ option is needed because you are on Mojave and latest Xcode, right?
Try exporting a CFLAGS environment variable containing that option before building uharfbuzz

jenskutilek:
Yes, I am on Mojave.

After installing the XCode command line tools, the build dependencies don't fail anymore, but building uharfbuzz still fails. How do I set the environment variable? If I try

```
CXXFLAGS='-std=libc++' pip install .
```

the compiler complains that the option is invalid and that valid options are `c++17` etc. (doesn't work either).

There is also a message that says XCode is not installed. Of course, XCode is installed.

And there is a recommendation to set the minimum deployment version to macos 10.9. How do I set the minimum macos version to 10.9? `CMAKE_OSX_DEPLOYMENT_TARGET=10.9` or ` MACOSX_DEPLOYMENT_TARGET=10.9`– no dice. 

Gotta love Apple's build system ...

PR #13, opened 2018-12-11
 remove use of legacy Py_UNICODE; use PEP393 unicode strings

anthrotype:
https://github.com/trufont/uharfbuzz/commit/ebdf61ee3d0a256cb8808219b32b493641e85115#r31635366

at least we have an excuse to explicitly drop support for python2.7


khaledhosny:
See https://github.com/trufont/uharfbuzz/commit/ebdf61ee3d0a256cb8808219b32b493641e85115#r31642745 for another suggestion.

anthrotype:
> hb_buffer_add_latin1, maybe we could use that if the string is of kind 1 but isn't pure ASCII.

correction: of course we can use `hb_buffer_add_latin1` every time the shortest-form canonical representation is Py_UCS1 (1 byte), since ASCII is a subset of latin1.

anthrotype:
@khaledhosny how about this?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -157,7 +168,7 @@ cdef class Buffer:
         if item_offset is None:
             item_offset = 0
         if item_length is None:
-            item_length = size
+            item_length = size - item_offset
```
khaledhosny: You can just pass `-1` here.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -176,35 +187,50 @@ cdef class Buffer:
...
```
khaledhosny: Same.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -176,35 +187,50 @@ cdef class Buffer:
...
         cdef char* cstr = text
         hb_buffer_add_utf8(
             self._hb_buffer, cstr, size, item_offset, item_length)
 
     def add_str(self, text: str,
                 item_offset: int = None, item_length: int = None) -> None:
```
khaledhosny: The default value for `item_offset` should be 0, and `item_length` should be -1, and then you won’t need any special handling if they were not passed.

khaledhosny:
Yes, I forgot about `_add_latin1()`, it was added for Chrome which I think uses similar string representation.

tests/test_uharfbuzz.py
```diff
@@ -73,7 +73,9 @@ def test_guess_set_segment_properties(self):
 
         assert buf.direction == "rtl"
         assert buf.script == "Hebr"
-        assert buf.language == "en-us"
+        # the guessed language seems to be locale specific
+        # assert buf.language == "en-us"
```
khaledhosny: Yes, HarfBuzz just the current locale for guessing the language (that is why `guess_set_segment_properties` should not be used in production code in general).

PR #24, opened 2020-01-20
return 0 (false) when the nominal glyph func returned .notdef

### LINKED ISSUE #23: custom glyph_h_advance_func + nominal_glyph_func causes problem with combining marks
I'm trying to use uharfbuzz with a custom advance width function.

The first thing I noticed that if you use `set_glyph_h_advance_func()` you _also_ have to use `set_nominal_glyph_func()`, which is a little inconvenient and unexpected, but not too hard to work around. So I also made a custom "nominal_glyph_func" function, that looks up a code point in the cmap as given by `fontTools`.

This works mostly alright.

The test font (included in the zip) contains a couple of precombined diacritics, and a mark feature that takes care of A + macroncmb and E + macroncmb. Without the custom functions, both precomposed and base + combining mark work as text input. But as soon as I use the custom functions neither work anymore. See the `showProblem` switch in the test script.

It may very well that I'm not using the API correctly, but if so, I don't have a clue what I'm doing wrong.

[testdata+script.zip](https://github.com/harfbuzz/uharfbuzz/files/4087466/testdata%2Bscript.zip)

justvanrossum:
Fixes #23.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -381,10 +381,8 @@ cdef hb_bool_t _nominal_glyph_func(hb_font_t* font, void* font_data,
     cdef Font py_font = <Font>font_data
     glyph[0] = (<FontFuncs>py_font.funcs)._nominal_glyph_func(
         py_font, unicode, <object>user_data)
-    if glyph[0]:
-        return 1
-    else:
-        return 0
+    # If the glyph is .notdef, return false, else return true
+    return int(glyph[0] != 0)
 
 
 cdef class FontFuncs:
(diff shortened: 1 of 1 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -381,7 +381,10 @@ cdef hb_bool_t _nominal_glyph_func(hb_font_t* font, void* font_data,
     cdef Font py_font = <Font>font_data
     glyph[0] = (<FontFuncs>py_font.funcs)._nominal_glyph_func(
         py_font, unicode, <object>user_data)
-    return 1
+    if glyph[0]:
+        return 1
+    else:
+        return 0
```
adrientetar: `return int(glyph[0] != 0)` ?

adrientetar: Also can you add just a one-line comment about the notdef thing before that

adrientetar:
Thanks!

PR #27, opened 2020-03-10
Add support for buffer message callbacks

simoncozens:


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -21,6 +21,10 @@ cdef extern from "Python.h":
     int PyUnicode_4BYTE_KIND
 
 
+cdef int msgcallback(hb_buffer_t *buffer, hb_font_t *font, const char* message, void* userdata):
+    (<object>userdata)(message)
+    return 1
+
```
justvanrossum: One more blank line here

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -220,6 +224,8 @@ cdef class Buffer:
     def guess_segment_properties(self) -> None:
         hb_buffer_guess_segment_properties(self._hb_buffer)
 
+    def set_message_func(self, realcallback) -> None:
+        hb_buffer_set_message_func(self._hb_buffer, msgcallback, <void*>realcallback, NULL)
```
justvanrossum: And one more blank line here

src/uharfbuzz/charfbuzz.pxd
```diff
@@ -114,6 +114,13 @@ cdef extern from "hb.h":
     hb_language_t hb_buffer_get_language(hb_buffer_t* buffer)
     void hb_buffer_set_language(hb_buffer_t* buffer, hb_language_t language)
     void hb_buffer_destroy(hb_buffer_t* buffer)
+    ctypedef hb_bool_t   (*hb_buffer_message_func_t) (hb_buffer_t *buffer,
```
justvanrossum: Have a look at similar declarations for formatting details, eg. line 150.

src/uharfbuzz/charfbuzz.pxd
```diff
@@ -114,6 +114,13 @@ cdef extern from "hb.h":
...
+                             hb_font_t   *font,
+                             const char  *message,
+                             void        *user_data)
+    void hb_buffer_set_message_func (hb_buffer_t *buffer,
```
justvanrossum: Same, have a look at line 85 for example

justvanrossum:
I would have loved to contribute a test, but the included test font has no OTL tables at all.

Maintainers: we could add a simple GSUB feature to the blank subset font. Would that be acceptable?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -220,6 +224,8 @@ cdef class Buffer:
     def guess_segment_properties(self) -> None:
         hb_buffer_guess_segment_properties(self._hb_buffer)
 
+    def set_message_func(self, realcallback) -> None:
```
adrientetar: Let's just call this callback

justvanrossum:
Wondering whether the msg string could be passed as an actual string instead of a bytes array?

anthrotype:
@simoncozens thanks for this!

> Maintainers: we could add a simple GSUB feature to the blank subset font. Would that be acceptable?

sgtm

justvanrossum:
I think this PR is good to go. I'll add a test case as a separate PR after the merge.

anthrotype:
@justvanrossum I added you as collaborator with write access, you should be able to merge PRs now.

justvanrossum:
Thanks! I wasn't sure whether I was supposed to, though, as a brand new committer.

PR #28, opened 2020-03-11
Message api tests

justvanrossum:
Adding tests for #27.


tests/test_uharfbuzz.py
```diff
@@ -166,3 +166,33 @@ def v_origin_func(font, gid, data):
...
+
+        def message(msg):
+            messages.append(msg)
+            infos_trace.append(buf.glyph_infos)
+            positions_trace.append(buf.glyph_positions)
+
+        buf.set_message_func(message)
+        hb.shape(blankfont, buf)
+        gids = [g.codepoint for g in buf.glyph_infos]
+        assert gids == [5, 4, 1, 2, 1]
```
anthrotype: it's not clear to what glyph these indexes are referring to (unless one inspects the test font). Can you maybe add some comments?

anthrotype:
@justvanrossum if you need a new release, it should be enough to push an annotated git tag in the form v{MAJOR}.{MINOR}.{PATCH} (e.g. `v0.7.0`) and the bots do the rest. Alternatively, you can also create a new release directly on Github (https://github.com/harfbuzz/uharfbuzz/releases/new), it should work the same.

justvanrossum:
Thanks, that's good to know!

PR #35, opened 2020-03-20
Add support to set script and language with OT tags instead of ISO and BCP 47

justvanrossum:
The current Buffer object has `script` and `language` properties, that use ISO-15924 and BCP 47 strings respectively.

However, when working directly with fonts, it is more convenient to use OT script and language tags.

I added this as setter methods, and not as properties, as it is not a 1-to-1 mapping.

This API allows me to fix justvanrossum/fontgoggles#68

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -166,13 +166,13 @@ cdef class Buffer:
         hb_buffer_set_script(
             self._hb_buffer, hb_script_from_string(cstr, -1))
 
-    def set_ot_language(self, value: str):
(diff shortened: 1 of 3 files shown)


justvanrossum:
Perhaps it would be better if the methods were called `set_ot_language_tag()` and `set_ot_script_tag()`.

justvanrossum:
Hm, since we're not "setting the OT tag", but "setting the language based on the OT tag", perhaps these names are best:
- `set_script_from_ot_tag(tag)`
- `set_language_from_ot_tag(tag)`

adrientetar:
Yeah, I was gonna say

PR #37, opened 2020-03-26
Reproduce crash on Windows?

justvanrossum:
I seem to get a crash on Windows if the original font data get gc-ed before the shaping is done. Let's see if AppVeyor can reproduce this.


justvanrossum:
The Travis failure is interesting: no crash, but wrong glyph IDs. So I think something is indeed fishy with gc-ed font data.

justvanrossum:
Ok, no crash on Windows on AppVeyor, yet the same wrong glyph IDs. Will commit a potential fix now.

PR #40, opened 2020-04-29
Add harfbuzz_version function; update to HB 2.6.5

josh-hadley:
- add `harfbuzz_function` function to top-level `uharfbuzz`
- add test for above
- update underlying HarfBuzz to 2.6.5

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,8 @@ project(uharfbuzz_skbuild)
 
 include(ExternalProject)
 
-set(HARFBUZZ_VERSION "2.6.4")
-set(HARFBUZZ_SHA256 "9413b8d96132d699687ef914ebb8c50440efc87b3f775d25856d7ec347c03c12")
+set(HARFBUZZ_VERSION "2.6.5")
+set(HARFBUZZ_SHA256 "126ea030bcb66fca518d43162443d337e60933c3729c23f8a25da3a1c7371dfd")
 
 # download, configure, build and install harbfuzz static library
 ExternalProject_Add(
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -26,6 +26,11 @@ cdef int msgcallback(hb_buffer_t *buffer, hb_font_t *font, const char* message, 
     return 1
 
 
+def harfbuzz_version() -> str:
+    cdef const char* cstr = hb_version_string()
+    cdef bytes packed = cstr
+    return packed.decode()
+
 cdef class GlyphInfo:
     cdef hb_glyph_info_t _hb_glyph_info
     # could maybe store Buffer to prevent GC
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -26,7 +26,7 @@ cdef int msgcallback(hb_buffer_t *buffer, hb_font_t *font, const char* message, 
     return 1
 
 
-def harfbuzz_version() -> str:
+def version_string() -> str:
     cdef const char* cstr = hb_version_string()
     cdef bytes packed = cstr
     return packed.decode()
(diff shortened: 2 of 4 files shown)


justvanrossum:
Thanks, looks good! Will merge and push a release once the tests finish.

anthrotype:
thanks @josh-hadley!

justvanrossum:
@anthrotype should the next release be 0.10.0, or should we bump to 1.0.0 perhaps?

behdad:
Why not `version_string()` instead of `harfbuzz_version()`?  The former matches the underlying API name.

anthrotype:
I don;t think we should limit ourselves to single digits, 10 is a valid minor version.
Another question is if we deem this to be a stable 1.0.0.
I'm ok if you want to set this as the 1.0.0

anthrotype:
good idea to match the hb API

josh-hadley:
> Why not `version_string()` instead of `harfbuzz_version()`? 

I wanted to make it clear that it was for the wrapped hb and not the version string for `uharfbuzz` itself.

anthrotype:
well, the wrapper already exposes a `__version__` attribute at top level like normally python modules do..
But I see your point. 🤔

justvanrossum:
As a compromise we could call it `hb_version_string`, although that's not consistent with how we wrap other parts of the API.

anthrotype:
as long as it's clearly documented `version_string` should not be a problem, IMO.
But if you prefer `hb_version_string`, go for it. Enough bikeshedding :)

justvanrossum:
Nothing is really documented in uharfbuzz... Either way: a doc string for this function would be nice.

josh-hadley:
@justvanrossum at any rate: whenever this gets merged/released, it would be good to get [fontgoggles](https://github.com/justvanrossum/fontgoggles) updated as well. HB 2.6.5 [fixes an issue with GPOS](https://github.com/harfbuzz/harfbuzz/pull/2087) which can be observed with [Adobe VF Prototype](https://github.com/adobe-fonts/adobe-variable-font-prototype/releases/tag/1.004). It manifests as "nothing in the GPOS" in FG.

So, `version_string` + docstring indicating that it's the HB version? LMK, I'll update soon.

anthrotype:
> version_string + docstring indicating that it's the HB version?

Sounds good to me. @justvanrossum ?

behdad:
I don't see the confusion you are trying to avoid.  `hb.version_string()` is clear to anyone using this.  `hb.hb_version_string()` doesn't offer any improvement.

josh-hadley:
> `hb.version_string()` is clear to anyone using this

Hope so, because it looks like [docstrings don't work in Cython](https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#docstrings) and this seems to be the case here, because a normal docstring doesn't show for `help()`. I also tried setting `__doc__="..."` on the function, but that causes `AttributeError: attribute '__doc__' of 'builtin_function_or_method' objects is not writable`. I guess that explains why there are no docstrings anywhere in `uharfbuzz` 😄 Please advise if there's a workaround...


Also, @anthrotype:
> well, the wrapper already exposes a `__version__` attribute at top level like normally python modules do..

Does it?
```
>>> import uharfbuzz as hb
>>> hb.__version__
  File "<stdin>", line 1, in <module>
AttributeError: module 'uharfbuzz' has no attribute '__version__'
>>> hb._harfbuzz.__version__
  File "<stdin>", line 1, in <module>
AttributeError: module 'uharfbuzz._harfbuzz' has no attribute '__version__'
```
Is it somewhere else?

anthrotype:
Ok, i thought we did expose that like we do elsewhere but apparently we don't. I guess we should. I can do that in a separate PR tomorrow.
Let's not split the hair here, either of the three proposed names works.

justvanrossum:
> HB 2.6.5 fixes an issue with GPOS which can be observed with Adobe VF Prototype. It manifests as "nothing in the GPOS" in FG.

Confirmed, and confirmed that this PR fixes that. Thanks! FG release will follow after uharfbuzz release, but I might wait for the `__version__` PR.

PR #42, opened 2020-05-08
Wrap hb_buffer_[s|g]et_cluster_level

khaledhosny:
Not sure what is the best way to handle enum values from user input, using int seems wrong as there is no error checking for bad values.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -1,4 +1,5 @@
 #cython: language_level=3
+from enum import IntEnum
 from charfbuzz cimport *
 from libc.stdlib cimport free, malloc
 from libc.string cimport const_char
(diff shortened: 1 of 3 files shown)


justvanrossum:
Would a Python Enum be usable here?

```python
from enum import IntEnum, auto


class ClusterLevel(IntEnum):
    HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES = 0
    HB_BUFFER_CLUSTER_LEVEL_MONOTONE_CHARACTERS = auto()
    HB_BUFFER_CLUSTER_LEVEL_CHARACTERS = auto()
    HB_BUFFER_CLUSTER_LEVEL_DEFAULT = auto()


print(ClusterLevel.HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES)
print(isinstance(ClusterLevel.HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES, int))
print(int(ClusterLevel.HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES))
print(int(ClusterLevel.HB_BUFFER_CLUSTER_LEVEL_MONOTONE_CHARACTERS))
print(int(ClusterLevel.HB_BUFFER_CLUSTER_LEVEL_CHARACTERS))
print(int(ClusterLevel.HB_BUFFER_CLUSTER_LEVEL_DEFAULT))
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -80,6 +81,13 @@ cdef class GlyphPosition:
         return self._hb_glyph_position.y_offset
 
 
+class BufferClusterLevel(IntEnum):
+    MONOTONE_GRAPHEMES = HB_BUFFER_CLUSTER_LEVEL_MONOTONE_GRAPHEMES
+    MONOTONE_CHARACTERS = HB_BUFFER_CLUSTER_LEVEL_MONOTONE_CHARACTERS
+    CHARACTERS = HB_BUFFER_CLUSTER_LEVEL_CHARACTERS
+    DEFAULT = HB_BUFFER_CLUSTER_LEVEL_DEFAULT
+
```
khaledhosny, line 89: I tried to move this class inside`Buffer` class below, so it does not need manual namespacing Buffer**.**ClusterLevel vs  BufferClusterLevel, but Cython does not allow that.

Issue #29, opened 2020-03-12
switch to meson build system


anthrotype:
upstream [Harfbuzz library will be switching to use meson](https://github.com/harfbuzz/harfbuzz/issues?q=is%3Aissue+is%3Aopen+label%3Ameson) as its build system, which means they may drop support for CMake some time in the future.
For uharfbuzz bindings, we use scikit-build as our python build system, which in turn uses CMake to orchestrate the build.
We should see if we can use Meson to build the cython extension module instead of CMake.

There's a package called [mesonpep517](https://gitlab.com/thiblahute/mesonpep517) that helps with creating python dists and wheels with meson (see [mesonpep517 docs](https://thiblahute.gitlab.io/mesonpep517/))

On mesonbuild repo there's an example of a cython module built with meson:
https://github.com/mesonbuild/meson/tree/master/test%20cases/python3/3%20cython

khaledhosny:
I think we should be able build the `src/harfbuzz.cc` file from HarfBuzz directly with setuptools and skip HarfBuzz build system altogether. Something like what we do in psautohint, since we don’t really need to detect any dependencies and all HarfBuzz optional features can be controlled with compiler definitions.

anthrotype:
> build the src/harfbuzz.cc file from HarfBuzz directly with setuptools and skip HarfBuzz build system

Yes, if we could do that, that'd be good.

> optional features can be controlled with compiler definitions

are those documented somewhere?

anthrotype:
you probably mean the HAVE_X, HAVE_Y, etc. sorts of preprocessor definitions that are controlled via harfbuzz's ./configure script.

I must say that the cmake-based setup that we currently have in uharfbuzz is quite neat though. I like that one pip install cmake and ninja and have cmake download tarball, verify, configure, install, compile, link, etc. But sure, I'm all for simplifying that if we can.

khaledhosny:
These are for external dependencies, but I was talking about https://github.com/harfbuzz/harfbuzz/blob/master/CONFIG.md#pre-defined-configurations

anthrotype:
Thank you Ebrahim?

justvanrossum:
I'm trying to build locally and am getting an error:

```
    src/uharfbuzz/_harfbuzz.cpp:646:10: fatal error: 'hb.h' file not found
```

What could be wrong on my end?

Full output:

```
(uharfbuzz-upstream) JvR:uharfbuzz just$ pip install -e .
[...]

PR #46, opened 2020-06-22
update to 2.6.8

justvanrossum:
This supersedes #44.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,8 @@ project(uharfbuzz_skbuild)
 
 include(ExternalProject)
 
-set(HARFBUZZ_VERSION "2.6.5")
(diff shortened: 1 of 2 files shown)


ebraminio:
Please consider using meson after the update also, #29, thanks :)

justvanrossum:
> Please consider using meson after the update also, #29, thanks :)

I know, but I don't know enough to make that happen myself, so here's to hoping that one of the other maintainers will have an opportunity to have a look at that...

anthrotype:
> Please consider using meson

Thanks Ebrahim for waiting a bit more before dropping cmake support. We'll probably end up building the sigle src/harfbuzz.cc file directly with setuptools and skip HarfBuzz build system altogether, like @khaledhosny has suggested. I need to find the time to do that soon.

ebraminio:
No problem, [harfbuzzjs](https://github.com/harfbuzz/harfbuzzjs/blob/master/build.sh) is using harfbuzz.cc (even CC src/hb*.cc works) from the beginning that may help you.

Issue #45, opened 2020-06-12
Option to link against installed harfbuzz


simoncozens:
My harfbuzz is often not like your harfbuzz, so downloading and building a vendor harfbuzz isn't something I want uharfbuzz to do. Obviously it's a good idea for platforms where you can't be sure harfbuzz is installed already, but it's unnecessary in many cases when you can just link against the system version. Can we make that an option?

anthrotype:
yes, we should add an option. Given we should soon switch away from cmake build system (see #29) it should be done as part of that effort.

ebraminio:
> yes, we should add an option. Given we should soon switch away from cmake build system (see #29) it should be done as part of that effort.

#49

BwL1289:
@anthrotype has there been any update on this? it doesn't look like the meson build system was ever adopted.

we need to link against our system harfbuzz to avoid breaking abi compatibility.

khaledhosny:
We currently use some experimental HarfBuzz APIs that are unlikely to be enabled in system HarfBuzz (because they are not stable APIs and HarfBuzz build does not enable them by default), so even if other technical issues get addressed, it might still not be possible to link to system HarfBuzz for most users.

BwL1289:
@khaledhosny thanks for the response. That will break ABI compatibility with a number of other libraries for us unfortunately.

Can you specify which experimental harfbuzz APIs are enabled in the vendored version?

khaledhosny:
We use `hb_subset_repack_or_fail()`.

BwL1289:
Thanks. We enable experimental APIs in our system harfbuzz so I might be able to fork as a last resort.

Would love to see an official option for this if possible.

khaledhosny:
PRs are always welcome.

BwL1289:
@khaledhosny our build is working on all platforms for CPython and fully working on windows but failing for PyPy for macos and ubuntu starting with PyPy 3.8. 

Are you able to help with getting PyPy to work? We are not familiar with it.

PR #48, opened 2020-06-22
Use harfbuzz's amalgam instead cmake

### LINKED ISSUE #29: switch to meson build system

ebraminio:
needs a `git clone https://github.com/harfbuzz/harfbuzz` on project's folder, we should find some idiomatic way for this, maybe git submodules?

fixes #29

self notes, put a `-fno-exceptions` as linkers might need it, find someway to add `HAVE_MMAP` and things on Unix.


ebraminio:
@justvanrossum @anthrotype can you help on progressing this? Thanks!

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1,10 +1,297 @@
...
 from enum import IntEnum
-from charfbuzz cimport *
 from libc.stdlib cimport free, malloc
 from libc.string cimport const_char
 from typing import Callable, Dict, List, Tuple
 
+from libc.stdint cimport uint8_t, uint16_t, uint32_t
+
+
+cdef extern from "hb.h":
```
ebraminio, line 10: Had to paste content of charfbuzz.pxd to this to make it work, thought `from .charfbuzz cimport *` would work but didn't? some help would be nice, thanks!

setup.py
```diff
@@ -35,21 +30,16 @@
     packages=["uharfbuzz"],
     zip_safe=False,
     setup_requires=["setuptools_scm"],
-    cmake_args=cmake_args,
     python_requires=">=3.5",
+    ext_modules = cythonize(Extension(
+        'uharfbuzz',
+        define_macros=[('HB_NO_MT', '1')],
```
ebraminio, line 36: Guess HB_NO_MT is no issue here? Some more common compile flags would be nice, like HAVE_ROUNDF and things, guess you can use HB_TINY and things also.

khaledhosny: No `HB_TINY` please, we want as much of HarfBuzz API and functionality as possible, size does not matter here.

ebraminio: Sure! Just wanted to indicate the place [your suggested configs](https://github.com/harfbuzz/uharfbuzz/issues/29#issuecomment-629475746) can be placed.

anthrotype: What does HB_NO_MT do? I suppose MT stands for "multi-threaded"? Why disabling that for the python extension?

src/uharfbuzz/_harfbuzz.pyx
anthrotype: `from charfbuzz cimport *` (without the `.`, as it was before) should work, does it not?

setup.py
khaledhosny: Yes, it disables multi-threading support in HarfBuzz (i.e. makes it multi thread unsafe). Do Python extensions get called from different threads simultaneously, if not then HarfBuzz does not not need to be multi thread safe.

src/uharfbuzz/_harfbuzz.pyx
ebraminio: It doesn't, can you help me and test it locally? The patch is of course just a PoC right now.

setup.py
ebraminio: > Why disabling that for the python extension?

This of course quickly is put together to show possibility of using amalgam for uharfbuzz, the reason that I've put it here is we should either declare we don't need MT support or we should detect which atomic/mutex things provider is available for the compile.

Using this simple to use python extension builder we can support #45 also just easily.

src/uharfbuzz/_harfbuzz.pyx
ebraminio: And not that I don't want fix it but I think at this point that it is a cython and it's unable to properly import things from the same directory and was thinking about moving the file to root of the project (not tested however if that help, will soon) and just thought as you are more familiar with things here and maybe interested on helping making it happen, if not, guess will be able to find my way somehow as it currently builds successfully at least.

pyproject.toml
```diff
@@ -5,7 +5,6 @@ requires = [
     "setuptools_scm >= 2.1",
     "cython >= 0.28.1",
     "scikit-build >= 0.9.0",
-    "cmake",
     "ninja",
```
ebraminio, line 8: no need for ninja also, will apply that also

setup.py
```diff
@@ -3,23 +3,18 @@
...
-for option in ("CYTHON_LINETRACE", "CYTHON_ANNOTATE"):
-    value = os.environ.get(option)
-    if value is not None and bool(int(value)):
-        cmake_args.append("-D{}:BOOL=TRUE".format(option))
-
-# On Mac, require OSX >= 10.9 so we can use te new libc++ targeting C++11
-if sys.platform == "darwin":
-    cmake_args.append("-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.9")
+extra_args = []
+if bool(int(os.environ.get('CYTHON_TRACE_NOGIL', '0'))):
```
ebraminio, line 16: Guess bool() isn't needed.

setup.py
```diff
@@ -3,23 +3,18 @@
...
 
-cmake_args = []
-for option in ("CYTHON_LINETRACE", "CYTHON_ANNOTATE"):
-    value = os.environ.get(option)
-    if value is not None and bool(int(value)):
-        cmake_args.append("-D{}:BOOL=TRUE".format(option))
-
-# On Mac, require OSX >= 10.9 so we can use te new libc++ targeting C++11
-if sys.platform == "darwin":
-    cmake_args.append("-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.9")
```
ebraminio, line 22: Wondering why this has anything about libc++ as harfbuzz won't use it anyway, of course we should make sure older compilers are using C++11 build flag but this isn't about it.

setup.py
```diff
@@ -35,21 +30,16 @@
...
+        'uharfbuzz',
+        define_macros=[('HB_NO_MT', '1')],
+        include_dirs=['harfbuzz/src'],
+        sources=['src/uharfbuzz/_harfbuzz.pyx', 'harfbuzz/src/harfbuzz.cc']
+    ), annotate=bool(int(os.environ.get('CYTHON_ANNOTATE', '0'))))
 )
 
 
 if __name__ == "__main__":
     import sys
```
ebraminio, line 44: ops, no need to

src/uharfbuzz/CMakeLists.txt
```diff
@@ -1,47 +0,0 @@
...
-    _harfbuzz PROPERTIES COMPILE_DEFINITIONS CYTHON_TRACE_NOGIL=1
-  )
-endif ()
-
-add_library(_harfbuzz MODULE ${_harfbuzz})
-
-if (APPLE)
-  find_library(APPLICATION_SERVICES_FRAMEWORK ApplicationServices)
-  if (NOT APPLICATION_SERVICES_FRAMEWORK)
-    message(FATAL_ERROR "HB_HAVE_CORETEXT was set, but we failed to find ApplicationServices, which is required")
```
ebraminio, line 38: is hb-coretext really used in uharfbuzz? easy to make it work just that I don't have mac for development and testing.

setup.py
```diff
@@ -3,23 +3,18 @@
...
-    value = os.environ.get(option)
-    if value is not None and bool(int(value)):
-        cmake_args.append("-D{}:BOOL=TRUE".format(option))
-
-# On Mac, require OSX >= 10.9 so we can use te new libc++ targeting C++11
-if sys.platform == "darwin":
-    cmake_args.append("-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.9")
+extra_args = []
+if bool(int(os.environ.get('CYTHON_TRACE_NOGIL', '0'))):
+    extra_args.append(('CYTHON_TRACE_NOGIL', '1'))
```
ebraminio, line 17: guess needs https://stackoverflow.com/a/28301932 also

src/uharfbuzz/CMakeLists.txt
khaledhosny: No it isn’t AFAIK, this was probably for the now gone `coretext_aat` shaper.

src/uharfbuzz/_harfbuzz.pyx
anthrotype: I think the issue is that you named the Extension just `uharfbuzz`, as standalone module without a package; before that, it was (implicitly) named `uharfbuzz._harfbuzz` (note the `uharfbuzz.` prefix, which is the name of the enclosing package where the extension module is located). Then cython compiler will extend the pxd include paths to also include the package directory (where charfbuzz.pxd is located).
Try to rename the Extension to `uharfbuzz._harfbuzz` and I think the `cimport` should work then.

setup.py
anthrotype: I wonder whether HB_NO_MT was defined in the old cmake based setup as well? Or was cmake detecting that automatically? I'm ok to disable multi-threading support in uharfbuzz until we an explicit need for it. Cython allows to [release/reacquire the GIL](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#releasing-the-gil) when running native code that doesn't interact with Python (so that other Python threads can run in parallel while native code is running). But you have to annotate that explicitly with `nogil` (we currently don't so we're holding the GIL and no MT is occurring).

setup.py
anthrotype: I am not sure actually. We can try to disable that and see if it works.

setup.py
anthrotype: Two environment variable need to be supported by setup.py: CYTHON_ANNOTATE and CYTHON_LINETRACE.

they are used here by the `tox` test runner to enable code coverage of cython files:
https://github.com/harfbuzz/uharfbuzz/blob/bcebfee9f659309ef1a1074ad090d41b86a03688/tox.ini#L16-L17

CYTHON_LINETRACE should, at the same time, pass linetrace=True in cythonize call, and define the CYTHON_TRACE_NOGIL=1 macro.

setup.py
ebraminio: > I wonder whether HB_NO_MT was defined in the old cmake based setup as well?

CMake was able to detect that so most likely it wasn't the case there. But it is really isn't that uncommon thing, Firefox https://searchfox.org/mozilla-central/source/gfx/harfbuzz/src/moz.build#112 still uses harfbuzz with that config and Chrome was used to use HB_NO_MT.

src/uharfbuzz/_harfbuzz.pyx
ebraminio: Thank you so much! It works now :)

setup.py
ebraminio: Thank you very much! `linetrace=True` results in failure so went with https://stackoverflow.com/a/28301932 solution.

Now I should learn how I can actually test this :)

anthrotype: I don't like that we override cython global defaults.. passing linetrace=True ought to work. What failure did you get?

anthrotype:
@ebraminio do you mind not rebase/force-pusing all the time, it makes it a bit harder to review. You can squash at the end after we LGTM

setup.py
ebraminio: Applying
```
diff --git a/setup.py b/setup.py
index d83e02b..052a9a6 100755
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ setup_params = dict(
     zip_safe=False,
     setup_requires=["setuptools_scm"],
     python_requires=">=3.5",
-    ext_modules = cythonize(extension, annotate=bool(int(os.environ.get('CYTHON_ANNOTATE', '0'))))
+    ext_modules = cythonize(extension, linetrace=True, annotate=bool(int(os.environ.get('CYTHON_ANNOTATE', '0'))))
 )
```

getting
```
  File "./setup.py", line 51, in <module>
    ext_modules = cythonize(extension, linetrace=True, annotate=bool(int(os.environ.get('CYTHON_ANNOTATE', '0'))))
  File "~/.local/lib/python3.8/site-packages/Cython/Build/Dependencies.py", line 961, in cythonize
    c_options = CompilationOptions(**options)
  File "~/.local/lib/python3.8/site-packages/Cython/Compiler/Main.py", line 563, in __init__
    raise ValueError(message)
ValueError: got unknown compilation option, please remove: linetrace
```

anthrotype: you need to pass `cythonize(..., compiler_directives={"linetrace": True})" as I do in skia-pathops here: 
https://github.com/fonttools/skia-pathops/blob/6581155e5151624ae9756392db025810933b74d6/setup.py#L108

ebraminio:
> @ebraminio do you mind not rebase/force-pusing all the time, it makes it a bit harder to review. You can squash at the end after we LGTM

Just days ago indicated to me by Behdad, OK :) I got that habit from gerrit that makes diff between amended commits also.

ebraminio:
So how I can test this in my system, CI configs aren't self explanatory, I am just currently issuing `./setup.py build` and not sure what should do next, without installing it on the entire system of course.

setup.py
ebraminio: Nice, done

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1,6 +1,6 @@
 #cython: language_level=3
 from enum import IntEnum
-from charfbuzz cimport *
+from .charfbuzz cimport *
```
ebraminio, line 3: Needed this

anthrotype:
to test on your system you run `tox`. By default running `tox` without passing `-e` option to select specific environments, or setting any `TOXENV` variable (which is the same as `-e`) you run all the envornments defined in the `tox.ini` default `envlist`. Currently `py3` (uses the current `python3` executable on PATH, does not collect coverage), `py3-cov` runs test with coverage, and finally the environment called (for lack of better word) `coverage`, which combines and creates the test coverage report.
If you only want to run a specific tox environment you do `tox -e py3`. Or `tox -e py3-cov,coverage` (comma separated list of tox environments run in sequence). If you do `tox -e py`, tox will use the same python where it was itself pip installed.
Yes, I know, this should be documented, but for python package maintainers `tox` is sort of familiar tool, used by many projects.

anthrotype:
if you don't want to use tox locally, you create the venv yourself, activate, install the project in editable or develop mode (`pip install -e .`, equivalent to `setup.py develop`), then whenever you make a change to the sources, make sure you run again `pip install -e .` so the extension is re-built and copied to the src/uharfbuzz dir. Or once you have installed in editable mode the first time, the subsequent times you need to rebuild you can directly run `python setup.py build_ext --inplace`, which will only rebuild the extension module and copy to destination.

anthrotype:
And to run the tests outside of tox, you just run `pytest` from within the venv where you have done `pip install -e .`.

ebraminio:
ci is now happy with this, can't believe it :)

setup.py
ebraminio: it targets C++11 anyway now.

ebraminio:
> And to run the tests outside of tox, you just run pytest from within the venv where you have done pip install -e ..

Thank you very much! :)

setup.py
```diff
@@ -22,7 +22,7 @@
     extension = Extension(
         'uharfbuzz._harfbuzz',
         libraries = ['harfbuzz'],
-        library_dirs = ['/usr/include/harfbuzz'],
+        include_dirs = ['/usr/include/harfbuzz'],
```
khaledhosny: This does not make sense, we shouldn’t be using headers from system. This should point to `harfbuzz/src` instead.

setup.py
```diff
@@ -3,23 +3,30 @@
...
+extension = Extension(
+    'uharfbuzz._harfbuzz',
+    define_macros=[('HB_NO_MT', '1')],
+    include_dirs=['harfbuzz/src'],
+    sources=['src/uharfbuzz/_harfbuzz.pyx', 'harfbuzz/src/harfbuzz.cc'],
+    language='c++',
+    extra_compile_args=[] if platform.system() == 'Windows' else ['-std=c++11'],
+)
 
 setup_params = dict(
```
khaledhosny, line 31:
```suggestion
setup(
```

And remove the `setup` call below.

ebraminio: done

ebraminio: oh sorry that I didn't what actually you wanted, applied.

setup.py
```diff
@@ -3,25 +3,32 @@
...
+    linetrace = True
+    extra_args.append(('CYTHON_TRACE_NOGIL', '1'))
+
+extension = Extension(
+    'uharfbuzz._harfbuzz',
+    define_macros=[('HB_NO_MT', '1')],
+    include_dirs=['harfbuzz/src'],
+    sources=['src/uharfbuzz/_harfbuzz.pyx', 'harfbuzz/src/harfbuzz.cc'],
+    language='c++',
+    extra_compile_args=[] if platform.system() == 'Windows' else ['-std=c++11'],
```
khaledhosny, line 28: The inline if condition is not the most readable. I’d move it outside similar to `extra_args`

setup.py
```diff
@@ -3,25 +3,32 @@
...
-cmake_args = []
-for option in ("CYTHON_LINETRACE", "CYTHON_ANNOTATE"):
-    value = os.environ.get(option)
-    if value is not None and bool(int(value)):
-        cmake_args.append("-D{}:BOOL=TRUE".format(option))
-
-# On Mac, require OSX >= 10.9 so we can use te new libc++ targeting C++11
-if sys.platform == "darwin":
-    cmake_args.append("-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.9")
+extra_args = []
```
khaledhosny, line 16: I don’t see where `extra_args` is used, or am I missing something?

setup.py
ebraminio: done

setup.py
ebraminio: oh, done

setup.py
```diff
@@ -19,13 +19,17 @@
     linetrace = True
     extra_args.append(('CYTHON_TRACE_NOGIL', '1'))
 
+extra_compile_args = []
+if platform.system() == 'Windows':
```
khaledhosny, line 23: Shouldn’t this be `!=` to actually match the old condition?

ebraminio: Oops, done

khaledhosny:
Not sure if this is possible, but I think the submodule should use the ref of the release tag not just the commit, so that is it more visible what release it is.

ebraminio:
> Not sure if this is possible, but I think the submodule should use the ref of the release tag not just the commit, so that is it more visible what release it is.

Makes sense, guess is done

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "harfbuzz"]
 	path = harfbuzz
 	url = https://github.com/harfbuzz/harfbuzz
+	branch = 2.6.8
```
khaledhosny, line 4: I don’t think this is what I meant, sorry for confusing you. I don’t think any action was needed here, what you had was fine.

ebraminio: no problem, gone

setup.py
```diff
@@ -3,25 +3,36 @@
...
-    cmake_args.append("-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.9")
+extra_args = []
+linetrace = False
+if int(os.environ.get('CYTHON_LINETRACE', '0')):
+    linetrace = True
+    extra_args.append(('CYTHON_TRACE_NOGIL', '1'))
+
+extra_compile_args = []
+if platform.system() != 'Windows':
+    extra_compile_args.append('-std=c++11')
```
ebraminio, line 24: what this means anyway? the issue is to make this work with older clang compilers preinstalled at macOS where they do support C++11 but aren't defaulting to it, Linux compilers (gcc/clang) were not the target but can benefit from it. This only isn't on Windows's cl.exe, msvc is targeting C++11 for a long time and harfbuzz doesn't support compile on older msvc compilers anyway, older mingw gcc compiler could benefit but detecting that is a bit tedious with low benefit so really not needed as all we wanted was to detect macos.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "harfbuzz"]
+	path = harfbuzz
+	url = https://github.com/harfbuzz/harfbuzz
```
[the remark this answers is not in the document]
ebraminio:
```
rm -rf harfbuzz && git submodule update --init --recursive && (cd harfbuzz && git status)
Submodule path 'harfbuzz': checked out 'e4203c14699cc0903a2b4611167a7b99532cfdb0'
HEAD detached at e4203c14
nothing to commit, working tree clean
```

but maybe adding `branch = 2.6.8` makes it safer or not, I can't assure that.

setup.py
```diff
@@ -3,25 +3,36 @@
...
-cmake_args = []
-for option in ("CYTHON_LINETRACE", "CYTHON_ANNOTATE"):
-    value = os.environ.get(option)
-    if value is not None and bool(int(value)):
-        cmake_args.append("-D{}:BOOL=TRUE".format(option))
-
-# On Mac, require OSX >= 10.9 so we can use te new libc++ targeting C++11
-if sys.platform == "darwin":
-    cmake_args.append("-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.9")
+extra_args = []
```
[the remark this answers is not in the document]
ebraminio: makes sense, done

PR #49, opened 2020-06-24
Add an option for linking against system library

### LINKED ISSUE #45: Option to link against installed harfbuzz

ebraminio:
Fixes #45, maybe can or should be default on Linux? Maybe not as that makes tests to fail depending on system version, some help on how to put it on CI would be nice, I am testing it locally using `USE_SYSTEM_HARFBUZZ=1 pip install . && pytest` which makes tests fail which is correct as my local harfbuzz installation isn't 2.6.8.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,24 +13,34 @@ here = os.path.abspath(os.path.dirname(__file__))
 with open(os.path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
-define_macros = [('HB_NO_MT', '1')]
+define_macros = []
 linetrace = False
 if int(os.environ.get('CYTHON_LINETRACE', '0')):
     linetrace = True
     define_macros.append(('CYTHON_TRACE_NOGIL', '1'))
 
 extra_compile_args = []
-if platform.system() != 'Windows':
-    extra_compile_args.append('-std=c++11')
 
-extension = Extension(
-    'uharfbuzz._harfbuzz',
-    define_macros=define_macros,
-    include_dirs=['harfbuzz/src'],
-    sources=['src/uharfbuzz/_harfbuzz.pyx', 'harfbuzz/src/harfbuzz.cc'],
-    language='c++',
-    extra_compile_args=extra_compile_args,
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -13,24 +13,34 @@
...
+        include_dirs=['harfbuzz/src'],
+        sources=['src/uharfbuzz/_harfbuzz.pyx', 'harfbuzz/src/harfbuzz.cc'],
+        language='c++',
+        extra_compile_args=extra_compile_args,
+    )
+else:
+    extension = Extension(
+        'uharfbuzz._harfbuzz',
+        define_macros=define_macros,
+        include_dirs=['/usr/include/harfbuzz'],
```
ebraminio, line 39: probably not cool as https://twitter.com/pcwalton/status/1272293882596700160 maybe I should ask `pkg-config`? not sure about the idiomatic here

ebraminio:
Some help I need, I've uninstalled uharfbuzz from pip and now I get `ImportError while importing test module '/home/ebrahim/Desktop/uharfbuzz/tests/test_uharfbuzz.py'.`

ebraminio:
Okay applied `USE_SYSTEM_HARFBUZZ=1 pip install .` and now `pytest` fails which is expected due to difference of system harfbuzz version with ones tests are against.

ebraminio:
PTAL

setup.py
khaledhosny: That wouldn’t work on any system with different HarfBiuzz installation path, and certainly wouldn’t help #45 case where HarfBuzz is most likely not installed there.

ebraminio: > where HarfBuzz is most likely not installed in this pat

oh, so let's close this till we get some idea how best we can solve that, thanks

PR #53, opened 2020-06-27
Expose and use HarfBuzz's mmap for font loading

### LINKED ISSUE #52: Use harfbuzz's mmap
uharfbuzz currently suggest reading fonts like, 
```
with open(sys.argv[1], 'rb') as fontfile:
    fontdata = fontfile.read()

text = sys.argv[2]

face = hb.Face(fontdata)
font = hb.Font(face)
upem = face.upem
```

but that's sub-optimal, one can enable HAVE_MMAP https://github.com/harfbuzz/harfbuzz/blob/master/src/hb-blob.cc#L478-L485 for non-Windows platform (Windows is already fine) and use `hb_blob_create_from_file` instead, I wonder however is uharfbuzz used in places that lots of fonts are loaded thus the it worth the font loading optimization.

ebraminio:
Fixes #52

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,6 +22,9 @@ if int(os.environ.get('CYTHON_LINETRACE', '0')):
 extra_compile_args = []
 if platform.system() != 'Windows':
     extra_compile_args.append('-std=c++11')
+    define_macros.append(('HAVE_MMAP', '1'))
+    define_macros.append(('HAVE_UNISTD_H', '1'))
+    define_macros.append(('HAVE_SYS_MMAN_H', '1'))
 
 extension = Extension(
     'uharfbuzz._harfbuzz',
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -328,6 +328,15 @@ cdef class Face:
         inst._reference_table_func = func
         return inst
 
+    @classmethod
+    def from_file_path(cls, file_path: str, int index=0):
+        cdef bytes path = file_path.encode()
+        cdef hb_blob_t* blob = hb_blob_create_from_file(<char*>path)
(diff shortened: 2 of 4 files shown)


ebraminio:
This one needs a little discussion for API naming and so, some feedback would be nice of course.

justvanrossum:
> This one needs a little discussion for API naming and so, some feedback would be nice of course.

I think `from_file_path` is fine. The PR looks good to me, but I'd prefer for someone with more knowledge have a look.

setup.py
```diff
@@ -22,6 +22,9 @@
 extra_compile_args = []
 if platform.system() != 'Windows':
```
ebraminio, line 23: Just to note Windows is already fine and able to find things needed for mmap.

ebraminio:
> I think `from_file_path` is fine. The PR looks good to me, but I'd prefer for someone with more knowledge have a look.

Sure :)

khaledhosny:
I’d rather we review the API and make it closer to the C API, e.g. make blob creation a separate step, create face only from blobs and deprecate passing data directly to it.

anthrotype:
> I’d rather we review the API and make it closer to the C API, e.g. make blob creation a separate step, create face only from blobs and deprecate passing data directly to it.

I agree. Can you file a separate issue for this?

khaledhosny:
> > I’d rather we review the API and make it closer to the C API, e.g. make blob creation a separate step, create face only from blobs and deprecate passing data directly to it.
> 
> I agree. Can you file a separate issue for this?

I’ll try. But since I’m slow to respond, please don’t let this block any ongoing development.

khaledhosny:
Sorry, closed by mistake.

khaledhosny:
Superseded by https://github.com/harfbuzz/uharfbuzz/pull/99

PR #54, opened 2020-06-27
Decrease blob refcount so can be freed when face is no longer used

ebraminio:
Otherwise won't be freed as far as I can see


ebraminio:
Guess it is safe to have this, right?

ebraminio:
Thank you very much :)

Issue #15, opened 2018-12-24
Higher level interface to uharfbuzz?


justvanrossum:
I've been writing a more Pythonic layer on top of uharfbuzz. I would gladly contribute it (I can see it as a submodule in uharfbuzz), if people here agree this is useful.

Just posting this here to test the waters, I can gladly make a PR:

[hbShape.py.zip](https://github.com/trufont/uharfbuzz/files/2707502/hbShape.py.zip)

Perhaps as something like `uharfbuzz.shape.Shaper`.

anthrotype:
sure, send a PR and we'll happily take a look :)

justvanrossum:
In the meantime I learned a lot more of the niceties that uharfbuzz offers and my higher level approach needs rethinking. What I've been using myself has evolved into something that is no longer general enough. Closing this, as I currently have no interest or time to work on it.

PR #56, opened 2020-07-22
Wrap (experimental) draw functions

simoncozens:
This will allow us to generate SVG (and do other drawing operations) from Python - see example in test suite.


tests/test_uharfbuzz.py
```diff
@@ -283,6 +300,29 @@ def test_message_func_crash(self, blankfont):
         hb.shape(blankfont, buf)
 
 
+    def test_draw_funcs(self, opensans):
+        funcs = hb.DrawFuncs.create()
```
anthrotype, line 304: we said `create()` classmethods are deprecated so it'd make sense to use default constructors in our own unit tests

simoncozens:
Yeah, it's pretty cool. :-) I'm already using it for doing rendering comparisons. See
https://github.com/simoncozens/font-engineering/blob/e6fd248c98b172528582be7341fd11cb89f72a61/vharfbuzz.py#L165-L169
```python
# v = Vharfbuzz("/Users/simon/Library/Fonts/SourceSansPro-Regular.otf")
# buf = v.shape("ABCj")
# svg = v.buf_to_svg(buf)
# import cairosvg
# cairosvg.svg2png(bytestring=svg, write_to="foo.png")
```

and https://github.com/simoncozens/font-engineering/blob/e6fd248c98b172528582be7341fd11cb89f72a61/hbuzzfuzz.py#L47-L55

```python
def visual_diff(s1, buf1, s2, buf2):
  png1 = cairosvg.svg2png(bytestring=s1.buf_to_svg(buf1))
  png2 = cairosvg.svg2png(bytestring=s2.buf_to_svg(buf2))
  im1 = imageio.imread(png1).sum(axis=2)
  im2 = imageio.imread(png2).sum(axis=2)
  padded = pad_max_shape([im1,im2])
  diff = padded[0]-padded[1]
  total_diff = np.sum(np.abs(diff))
  return (total_diff / (im1.shape[0] * im1.shape[1]) * 100)
```

tests/test_uharfbuzz.py
justvanrossum: Why even add a deprecated method to a new object?

anthrotype:
nice! 

> vharfbuzz

LOL. Maybe we could add a nicer more pythonic API to uharfbuzz on top of a low-level one that mimics the C API.

tests/test_uharfbuzz.py
simoncozens: Probably the honest answer is "because I don't *really* know what I'm doing with Cython so I'm just copy-and-pasting similar code and then hacking it until it works." Anyway, will fix.

tests/test_uharfbuzz.py
```diff
@@ -283,6 +300,29 @@ def test_message_func_crash(self, blankfont):
...
+        container = { "output_string": "" }
```
justvanrossum, line 305: I would just use a list here, and `container.append(...)` in the callbacks. Then you could do `output_string = "".join(container)`, although for the sake of the test I think the fact that this is SVG is not relevant.

tests/test_uharfbuzz.py
justvanrossum: Wait, that is my mode of operation here, too :)

simoncozens:
> LOL. Maybe we could add a nicer more pythonic API to uharfbuzz on top of a low-level one that mimics the C API.

Well, this was discussed in #15. You're welcome to take what I've got in Vharfbuzz and use that. I would offer a PR but I'm not sure how to organise the code when you have both Cython and Python parts to a module.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -690,3 +690,131 @@ def ot_layout_get_baseline(font: Font,
...
+        return self._quadratic_to_func
+
+    def close_path_func(self):
+        return self._close_path_func
+
+    def user_data(self):
+        return self._user_data
+
+    def set_move_to_func(self,
+                                 func: Callable[[
```
justvanrossum, line 769: Something is off with the alignment of these statements.

justvanrossum:
Higher level: I think this (outline extraction) functionality would be immediately much more useful if there was a function that would draw to a fonttools pen. Something like this:

```python
    def test_draw_glyph_to_pen(self, opensans):
        from fontTools.pens.svgPathPen import SVGPathPen
        pen = SVGPathPen(None)
        hb.draw_glyph_to_pen(opensans, 1, pen)
        assert(pen.getCommands() == "M1120,0L938,465L352,465L172,0L0,0L578,1468L721,1468L1296,0L1120,0ZM885,618L715,1071Q682,1157 647,1282Q625,1186 584,1071L412,618L885,618Z")
```

tests/test_uharfbuzz.py
```diff
@@ -283,6 +300,29 @@ def test_message_func_crash(self, blankfont):
...
+        def close_path(c):
+            c.append("Z")
+
+        funcs.set_move_to_func(move_to)
+        funcs.set_line_to_func(line_to)
+        funcs.set_cubic_to_func(cubic_to)
+        funcs.set_quadratic_to_func(quadratic_to)
+        funcs.set_close_path_func(close_path)
+        funcs.draw_glyph(opensans, 1, container)
+        assert("".join(container) == "M1120,0L938,465L352,465L172,0L0,0L578,1468L721,1468L1296,0L1120,0ZM885,618L715,1071Q682,1157 647,1282Q625,1186 584,1071L412,618L885,618Z")
```
justvanrossum, line 323: `assert x == y` does not need parentheses.

simoncozens:
I agree that a pen writer would be useful. At the moment, uharfbuzz does very little other than wrapping the C API functions, so it would be quite a different thing to add a function like that.

I'll add it to vharfbuzz for now, until we've agreed what level uharfbuzz should be pitched at.

justvanrossum:
I don't see uharfbuzz as a super pure wrapper. It does for example add quite some object orientedness on top of the hb API. 

Creating an object and having to register a handful of callbacks explicitly isn't all that different from passing an object that defines said handful of callbacks as methods. Except the latter is almost infinitely more usable from a Python perspective. To me an obvious case of "practicality beats purity".

simoncozens:
Actually it turned out not to be too nasty. :-)

justvanrossum:
Awesome, thanks!

simoncozens:
And now fontTools becomes a dependency for the tests. Urgh.

justvanrossum:
Ah ugh, I somehow mistakenly thought the test suite already depended on fonttools. It needs to be added to reuirements-dev.txt.

justvanrossum:
Urgh, duplicate comment.

justvanrossum:
Instead of a fonttools dependency, we could add a test pen, say something like this (simplified from RecordingPen):

```python
class TestPen:
    def __init__(self):
        self.value = []
    def moveTo(self, p0):
        self.value.append(('moveTo', (p0,)))
    def lineTo(self, p1):
        self.value.append(('lineTo', (p1,)))
    def qCurveTo(self, *points):
        self.value.append(('qCurveTo', points))
    def curveTo(self, *points):
        self.value.append(('curveTo', points))
    def closePath(self):
        self.value.append(('closePath', ()))
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -420,6 +420,26 @@ cdef class Font:
         packed = name
         return packed.decode()
 
+    def draw_glyph_to_pen(self, gid: int, pen):
```
anthrotype, line 423: Usually one draws a glyph _with_ a pen, not _to_. I would have called this method `draw_glyph_with_pen`

Also, do we need to create a new DrawFuncs() every time one calls draw_glyph_with_pen? Can't we create one for each Font instance and reuse?

justvanrossum: > Usually one draws a glyph with a pen, not to. I would have called this method draw_glyph_with_pen

True. The whole "pen" thing in my mind has become so abstract that I don't think about it that way anymore. (Just like I don't associate the computer concept of "file" with a physical file to keep papers in.)

That said, I won't mind if this were changed. We should decide before the next release, though, as this PR has already been merged. (Speaking of which: I don't know how to update the harfbuzz submodule to the latest release.)

> Also, do we need to create a new DrawFuncs() every time one calls draw_glyph_with_pen? Can't we create one for each Font instance and reuse?

That is a good point, but could be implemented later as an optimization.

anthrotype: > I don't know how to update the harfbuzz submodule to the latest release

you `cd` into the submodule directory, git pull and then checkout the tagged commit, then cd .. back to the root of the parent repo and git add and git commit the change

justvanrossum: @simoncozens, would you like to revisit the naming of this method, or is this good to go?

simoncozens: I’m happy with it myself, but we should probably rename it for other people’s sanity. :-) will send a PR in the morning.

behdad: > draw_glyph_with_pen

FWIW what I expected was just `draw_glyph` that would then probe pen to see what protocol to talk to it; ie. duck-typing method-overloading.

PR #60, opened 2020-09-11
Install cython in CI environment before attempting source distribution

### LINKED ISSUE #59: Source ZIP didn't get uploaded to PyPi for 0.13.0
Trying to update Arch Linux packaging, and the source ... isn't there yet. The release and a bunch of prebuilt wheels are, but not the source ZIP that was previously available on releases. It's possible to switch to Github sources, but the use of `setuptools-scm` makes that complicated and I would have to setup a patch to remove it and fill in the missing release info. The generated source dist is much easier to work with.

alerque:
Closes #59

My understanding from the discussion there is that installing setuptools by hand is unnecessary, but than installing cython is.

I don't really have a way to test that this is going to work, so please review carefully @anthrotype!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ branches:
 install:
   - $PYTHON -m pip install $PIP_INSTALL_OPTIONS pip
   - $PYTHON -m pip install $PIP_INSTALL_OPTIONS cibuildwheel==1.1.0
+  - $PYTHON -m pip install $PIP_INSTALL_OPTIONS setuptools_scm
 
 script:
   - $PYTHON -m cibuildwheel --output-dir wheelhouse
@@ -39,7 +40,6 @@ script:
     if [[ $TRAVIS_TAG ]]; then
       $PYTHON -m pip install $PIP_INSTALL_OPTIONS twine
       if [[ $TRAVIS_OS_NAME == osx ]]; then
-        $PYTHON -m pip install $PIP_INSTALL_OPTIONS setuptools
         $PYTHON setup.py sdist
         $PYTHON -m twine upload dist/*.zip
       fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,7 @@ script:
     if [[ $TRAVIS_TAG ]]; then
       $PYTHON -m pip install $PIP_INSTALL_OPTIONS twine
       if [[ $TRAVIS_OS_NAME == osx ]]; then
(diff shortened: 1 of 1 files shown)


anthrotype (commented):
I suggest you keep upgrading setuptoools as it's always better to use latest when doing packaging (rather than whatever comes with given python distribution)

justvanrossum:
I think what @anthrotype suggested is to keep the install setuptools line, and only add the install cython line.

alerque:
@justvanrossum I did keep it, but if his line of reasoning ("better to use the latest when doing packaging") makes sense for the source distribution then it probably makes sense for the built distribution too. There doesn't seem to be any sense in posting a release where some of the builds are generated with one version of tooling and some another. Hence I moved the line up to cover the built distributions as well as the source one.

justvanrossum:
But the wheels all have already been built at this stage. All that's happening (in the non-osx case) is uploading the wheels.

alerque:
Roger that. :facepalm: I've now moved it up to where it will actually do what the rational above suggested it should.

justvanrossum:
I don't think that install setuptools_scm line is needed at all, but I'll defer to @anthrotype's judgement.

alerque:
I'd be in favor of taking it out too if there is not a known issue with the version running in Travis's Python runner by default. Then again `pip` is also getting upgraded there, so if that's being done maybe it does make sense. I agree this is Cosimo's call, I'm not a Pythonista.

justvanrossum:
But the "install setuptools" line should stay in the osx section.

alerque:
> But the "install setuptools" line should stay in the osx section.

I'm sorry I don't see the logic for this at all. The only reason we're keeping it at all is because Cosimo suggested above that "it's always better to use latest when doing packaging". Hence if it makes any sense at all to keep, it makes sense to move before we do packaging rather than half way through the process. It needs to be before the `cibuildwheel` step that does the other half of the packaging, upgrading setuptools just for the `sdist` step doesn't make any sense to me. Why do you say it should stay there?

justvanrossum:
Solving this in master now, CI started for 0.13.1.

justvanrossum:
The problem is a missing cython dependency, so let's solve that. (Just done on master)

justvanrossum:
Closing this in favor of 3c1ef40b80590e0a0506180d8a0bbc09355135ec. Apologies if this seems rushed.

Issue #64, opened 2020-09-19
(Un)setting features for part of a buffer: what could that look like?


justvanrossum:
We currently don't support setting or unsetting of features for only part of the input buffer.

I don't yet understand the underlying HB calls needed to achieve this, and I also don't yet see what such an API could look like from the Python side. Could it be an operation on the buffer?

behdad:
`hb_shape()` API, just set `hb_feature_t`. The `start`/`end` ranges in `hb_feature_t` refer to cluster values. Depends on how you are converting Python string to add to HB buffer. If you do it this way:

https://github.com/harfbuzz/harfbuzz/blob/master/src/sample.py#L30

Then you can use Python string indices as your `start`/`end`. A value of 0 turns feature off.

khaledhosny:
Currently uharfbuzz’s `shape()` takes features as a dictionary of `{tag: bool}` which is unfortunate, but it uses `hb_feature_from_string()` to parse the tag part, so you can workaround this by using the extended, python-like syntax for the tag (`"tag[start:end]"`) and it should achieve the same as the C API.

khaledhosny:
e.g.:
```Python
shape(font, buffer, {"dlig[3:5]": True, "hlig[0:7]": True, "liga[0:7]": False})`
```

justvanrossum:
Oh wow, it pretty much already works, I had no idea! Thanks.

I think two things need to change:

- the bool dict value is already interpreted as an int: the features typedef should be changed accordingly
- Instead of a dict, it should also accept a list of tuples

I'll try to prepare a PR later this week.

justvanrossum:
Before:
`features: Dict[str, bool] = None`
After:
`features: Union[Dict[str, int], List[Tuple[str, int]]] = None`

behdad:
By `List[Tuple[str, int]]]` you mean providing the features for *each* character?

I advise you to look more into the index/cluster values. Because if they don't match between Python string indices and what is passed to HarfBuzz it will be a problem. See my other comment. The `uharfbuzz` code uses different (utf8/utf16/codepoints) API depending on internal CPython storage of the string, NOT the user-facing indices in the string. Is that a problem? Probably not. But want to make sure. Do we not care about 2byte Python runtime anymore? If that's true, the code might be okay.

I advise you also look into using Python `Slice` objects if that simplifies things. And you probably don't need to use `hb_feature_from_string()` anyway.

justvanrossum:
[...]

PR #65, opened 2020-09-21
Implement a more Pythonic feature slice notation

justvanrossum:
As discussed in #64.

The test case is a little overcomplicated because the only GSUB feature available is one `calt` feature.

This does _not_ remove the existing `features={"aalt[2]": 3}` notation, as that would not be backwards compatible. (I hadn't realized this worked at all until now, so I don't personally care for that notation.)

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -588,7 +588,7 @@ def shape(font: Font, buffer: Buffer,
     cdef char* cstr
     cdef hb_feature_t feat
     cdef const char **c_shapers
-    if features is None:
+    if not features:
         size = 0
         hb_features = NULL
     else:
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -3,7 +3,7 @@ from enum import IntEnum
 from .charfbuzz cimport *
 from libc.stdlib cimport free, malloc
 from libc.string cimport const_char
-from typing import Callable, Dict, List, Tuple
+from typing import Callable, Dict, List, Sequence, Tuple, Union
 
 
 cdef extern from "Python.h":
@@ -579,7 +579,8 @@ cdef const char ** to_cstring_array(list_str):
     return ret
 
 
-def shape(font: Font, buffer: Buffer, features: Dict[str, bool] = None,
+def shape(font: Font, buffer: Buffer,
+        features: Dict[str,Union[int,bool,Sequence[Tuple[int,int,Union[int,bool]]]]] = None,
         shapers: List[str] = None) -> None:
     cdef unsigned int size
     cdef hb_feature_t* hb_features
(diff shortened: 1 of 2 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -579,26 +579,43 @@ cdef const char ** to_cstring_array(list_str):
...
+            if isinstance(value, int):
+                feat.value = value
+                hb_features[i] = feat
+                i += 1
+            else:
+                for start, end, value in value:
+                    feat.value = value
+                    feat.start = start
+                    feat.end = end
+                    hb_features[i] = feat
```
khaledhosny, line 617: Hmm, is this re-use of feat really “safe”, shouldn’t you be declaring it inside the loop, or otherwise making a copy (this C but using Python syntax really makes me nervous). Also in this branch `hb_feature_from_string()` is superfous and we probably don’t want to allow the extended syntax for the tag her, so `feat.tag = hb_tag_from_string()` might be better.

justvanrossum: Yes it is safe, it really is C semantics with Python syntax. The `hb_features[i] = feat` really copies the value into the array — it is not a pointer. (There is a single `hb_feature_t feat` variable that is used over and over again, this is not new with this PR.)

I agree that `hb_feature_from_string()` would be superfluous, but users may depend on this behavior, so it would not be a backwards compatible change.

khaledhosny: > Yes it is safe, it really is C semantics with Python syntax. The `hb_features[i] = feat` really copies the value into the array — it is not a pointer. (There is a single `hb_feature_t feat` variable that is used over and over again, this is not new with this PR.)

I see. Sorry for the noise.

> I agree that `hb_feature_from_string()` would be superfluous, but users may depend on this behavior, so it would not be a backwards compatible change.

I mean only for the tuple value case (there is no backward compatability here) since accepting `{"aalt[2:4]": (3, 0, 5)}` is rather weird.

khaledhosny (approved):

src/uharfbuzz/_harfbuzz.pyx
justvanrossum: > I mean only for the tuple value case (there is no backward compatability here) since accepting {"aalt[2:4]": (3, 0, 5)} is rather weird.

Ah, of course you're right. What would be the best/easiest way to get a `hb_tag_t` from a tag string?

khaledhosny: `hb_tag_from_string()`

justvanrossum: (Sorry for such a basic question, I had a bit of a blackout, could have found it myself easily.)

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -604,12 +604,13 @@ def shape(font: Font, buffer: Buffer,
             assert i < size, "index out of range for feature array capacity"
             packed = name.encode()
             cstr = packed
```
khaledhosny, line 606: While at it, this `cstr` seems unused.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -579,34 +579,50 @@ cdef const char ** to_cstring_array(list_str):
...
+                    size += 1
+                else:
+                    size += len(value)
+            hb_features = <hb_feature_t*>malloc(size * sizeof(hb_feature_t))
+            i = 0
+            for name, value in features.items():
+                assert i < size, "index out of range for feature array capacity"
+                packed = name.encode()
+                if isinstance(value, int):
+                    hb_feature_from_string(packed, len(packed), &feat)
```
behdad, line 605: Please change this to `hb_tag_from_string()` like below as well.

justvanrossum: That would make this change non-backwards compatible, are we ok with that?

behdad: Intresting. *If* we concluded the range numbers work fine (which looks like they do), then I'm agree keeping it is fine.

justvanrossum: Not sure what you mean. I think the range numbers are fine.

But before this PR, the `features` dict already supported a notation like `features={"aalt[2:3]", 2}`. That feels accidental to me (I didn't even realized it did that until Khaled pointed it out), but perhaps @adrientetar built it this way on purpose, and perhaps there are users out there that use it.

behdad: Yes yes. I understand. And I said I agree with your assessment. lgtm.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -579,34 +579,50 @@ cdef const char ** to_cstring_array(list_str):
     return ret
 
 
-def shape(font: Font, buffer: Buffer, features: Dict[str, bool] = None,
+def shape(font: Font, buffer: Buffer,
+        features: Dict[str,Union[int,bool,Sequence[Tuple[int,int,Union[int,bool]]]]] = None,
```
adrientetar, line 583: nit: you can use type alias to make this less stuffed

Issue #66, opened 2021-02-02
Readme example: `hb.ot_font_set_funcs(font)` needed?


madig:
https://harfbuzz.github.io/harfbuzz-hb-ot-font.html#hb-ot-font-set-funcs says it's already called on `hb_font_create`, which `uharfbuzz.Font` does. Does it need to be there?

justvanrossum:
I think you're right and that it's indeed redundant. Care to make a PR?

PR #67, opened 2021-02-03
Remove `hb.ot_font_set_funcs(font)` from example

madig:
Not needed, as `uharfbuzz.Font` does it implicitly via `hb_font_create`. Also trim some lines from the example.

Closes https://github.com/harfbuzz/uharfbuzz/issues/66.


justvanrossum:
Thanks!

PR #80, opened 2021-05-07
Add font.get_glyph_extents() wrapper for hb_font_get_glyph_extents()

justvanrossum:


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -414,6 +414,16 @@ cdef class Font:
         else:
             return None
 
+    def get_glyph_extents(self, gid: int):
+        cdef hb_glyph_extents_t extents
+        success = hb_font_get_glyph_extents(self._hb_font, gid, &extents)
+        if success:
+            return (extents.x_bearing, extents.y_bearing,
+                    extents.width, extents.height)
```
khaledhosny, line 422: It would have nicer if the function returned a namedtuple, having to guess which value is which Everytime reading code using this function is a bit unfortunate.

justvanrossum: You're right, will fix.

PR #81, opened 2021-05-08
Add misc font wrappers

### LINKED ISSUE #22: glyph indexes
uharfbuzz has only support to add glyphs to a buffer as text or as a list of unicodes: `add_codepoints`, `add_str`, `add_utf8` but not as glyph indexes this would be super handy... 

I dont know if harfbuzz its self has support for this

see https://github.com/typemytype/featurePreviewRoboFontExtension/issues/8

justvanrossum:
Wrap:
- hb_font_get_nominal_glyph (this should resolve #22)
- hb_font_get_var_coords_normalized
- hb_font_set_var_coords_normalized

And:
- add subset of MutatorSans so we have a small VF to test with


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -423,6 +423,30 @@ cdef class Font:
...
+    def get_nominal_glyph(self, unicode: int):
+        cdef hb_codepoint_t gid
+        success = hb_font_get_nominal_glyph(self._hb_font, unicode, &gid)
+        return gid if success else None
+
+    def get_var_coords_normalized(self):
+        cdef unsigned int length
+        cdef const int *coords
+        coords = hb_font_get_var_coords_normalized(self._hb_font, &length)
+        return [coords[i] / 0x4000 for i in range(length)]
```
khaledhosny, line 435: Probably it is just my ignorance, but it is not clear from the code what this division is about (and the multiplication below), so a comment might help.

justvanrossum: I'll add a comment. It converts 2.14 to float.

justvanrossum: Done in master: 2f533d9d1131a2e630dfbe71956da6816db05785

khaledhosny: Thanks!

Issue #83, opened 2021-05-16
coretext shaper causes segfault on "sparse" font


justvanrossum:
I have a font that only has GPOS, cmap, head, maxp and post tables, which works perfectly fine for the default (ot) shaper, but crashes hard when specifying the coretext shaper on macOS.

Relates to https://github.com/justvanrossum/fontgoggles/pull/144

(FontGoggles compiles this kind of minimal fonts on the fly to do shaping for UFO sources: metrics come from the source via hb callbacks, and outlines are directly drawn from the sources, too.)

I'll open a PR that reproduces the problem.

behdad:
> (FontGoggles compiles this kind of minimal fonts on the fly to do shaping for UFO sources: metrics come from the source via hb callbacks, and outlines are directly drawn from the sources, too.)

Beautiful. I had designed HB to be used that way :).

Does the crash reproduce with hb-view?

justvanrossum:
> Beautiful. I had designed HB to be used that way :).

It's pretty fantastic indeed :)

> Does the crash reproduce with hb-view?

It fails gracefully like this:
```
$ hb-view testfont.ttf ABC --shapers=coretext
hb-view: all shapers failed.
hb-view: FT_New_Memory_Face fail
```
So no, it doesn't crash.

PR #84, opened 2021-05-16
Reproducing and fixing crash of #83

justvanrossum:
Reproduces  #83


justvanrossum:
I notice that the return value of `hb_shape_full`  is not checked, but that doesn't immediately explain the crash to me.

https://github.com/harfbuzz/uharfbuzz/blob/2f533d9d1131a2e630dfbe71956da6816db05785/src/uharfbuzz/_harfbuzz.pyx#L665-L667

hb-shape reports that "all shapers failed", so I'm fairly sure `hb_shape_full` returns false here.

justvanrossum:
(The GH CI action is queued, and somehow it's been stuck for more than 20 minutes. Fingers crossed it will run eventually.)

justvanrossum:
Here's a relevant bit from the crash report, in case that helps anyone figure out what we do wrong:
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff6825933a __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff68315e60 pthread_kill + 430
2   libsystem_c.dylib             	0x00007fff6817093e raise + 26
3   libsystem_platform.dylib      	0x00007fff6830a5fd _sigtramp + 29
4   ???                           	000000000000000000 0 + 0
5   _harfbuzz.cpython-37m-darwin.so	0x000000010abb7751 hb_shape_plan_key_t::init(bool, hb_face_t*, hb_segment_properties_t const*, hb_feature_t const*, unsigned int, int const*, unsigned int, char const* const*) + 881 (hb-shaper-list.hh:42)
6   _harfbuzz.cpython-37m-darwin.so	0x000000010abba356 hb_shape_plan_create_cached2 + 134 (hb-shape-plan.cc:537)
7   _harfbuzz.cpython-37m-darwin.so	0x000000010abba726 hb_shape_full + 54 (hb-shape.cc:135)
8   _harfbuzz.cpython-37m-darwin.so	0x000000010ab6de51 __pyx_pw_9uharfbuzz_9_harfbuzz_3shape(_object*, _object*, _object*) + 1217 (_harfbuzz.cpp:11782)
9   org.python.python             	0x000000010a2aa3fb _PyMethodDef_RawFastCallKeywords + 379
```

simoncozens:
What if you supply an OS2 table? I’ve had problems with CoreText and fonts without one.

justvanrossum:
> What if you supply an OS2 table? I’ve had problems with CoreText and fonts without one.

Thanks, I will try that.

My main concern, however, is that it crashes. I'm fine with uhb/hb/coretext not being able to work with this font, but it should do so gracefully. Like hb-shape: it doesn't provide useful output, but shows an error message and doesn't crash.

justvanrossum:
FWIW: adding an OS/2 table doesn't make the crash go away.

khaledhosny:
If I add `fallback` to the list of shapers, it does not crash. Something is not handling well the fact that al shapers failed.

khaledhosny:
`hb_shape()`’s `shapers` argument is a `NULL` terminated array, but uharfbuzz is not `NULL`-terminating it. With the following patch all tests pass:
```diff
diff --git a/src/uharfbuzz/_harfbuzz.pyx b/src/uharfbuzz/_harfbuzz.pyx
index cebbe75..c6c8a34 100644
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -620,9 +620,10 @@ cdef class FontFuncs:


 cdef const char ** to_cstring_array(list_str):
-    cdef const char **ret = <const char **>malloc(len(list_str) * sizeof(char *))
+    cdef const char **ret = <const char **>malloc((len(list_str) + 1) * sizeof(char *))
     for i in range(len(list_str)):
         ret[i] = PyUnicode_AsUTF8(list_str[i])
+    ret[i + 1] = NULL
     return ret
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -619,21 +618,14 @@ cdef class FontFuncs:
...
-
 def shape(font: Font, buffer: Buffer,
         features: Dict[str,Union[int,bool,Sequence[Tuple[int,int,Union[int,bool]]]]] = None,
         shapers: List[str] = None) -> None:
     cdef unsigned int size
     cdef hb_feature_t* hb_features
     cdef bytes packed
     cdef hb_feature_t feat
-    cdef const char **c_shapers
+    cdef const char* c_shapers[10]
```
justvanrossum, line 628: Is this WIP or is this a revision on your proposed fix from https://github.com/harfbuzz/uharfbuzz/pull/84#issuecomment-885965144

justvanrossum: Sorry, I only now saw the commit msg! If you think 9 is a good number, let's go with this!

justvanrossum:
I can't approve this PR (as I'm the initiator) but it looks good to me. Is it ready to merge?

anthrotype (approved):

src/uharfbuzz/_harfbuzz.pyx
khaledhosny: It should be plenty, HarfBuzz currently has 6 shapers and the number is expected to go down not up.

Issue #86, opened 2021-05-18
No win_amd64 wheels were produced for last release (0.16.0)


justvanrossum:
https://pypi.org/project/uharfbuzz/0.16.0/#files

anthrotype:
we run cibuildwheel to build both the win32 adn amd64 wheel one after the other in the same windows instance. Something is leaving behind the repository in dirty state because by the time the 64-bit wheel gets built, setuptools_scm sees untracked files and assigns the new wheel a local version which can't be uplaoded to PyPI

https://ci.appveyor.com/project/fonttools/uharfbuzz/builds/39074250#L1916

anthrotype:
it's strange because it was working fine untile the penultimate release

anthrotype:
oh! I see what's going on.. After running `pytest`, I see the `harfbuzz` git submodule contains untracked content, it's  `test/subset/__pycache__`. The fix should be to add that to upstream harfbuzz .gitignore. But why does that get imported in our tests?

justvanrossum:
The other strange thing is that I didn't get any AppVeyor fail notification after I tagged a release. Is that to be expected, with the way AppVeyor is set up?

(I only accidentally discovered there were no 64 bit wheels because I saw another CI process suspiciously build uharfbuzz wheels...)

anthrotype:
I just set it up to email fonttools-dev@googlegroups.com when build fails or build status change

anthrotype:
the appveyor bot is under the fonttools organization, should we spam harfbuzz instead? It doesn't really matter

justvanrossum:
I don't really know. Longer-term I'd prefer to move away from AppVeyor and have #87 working instead.

anthrotype:
if you like i can invite you as collaborator to the appveyor fonttools account so you should get email notifications (need an email address to send to)

justvanrossum:
Sure, my GH username @ gmail is good.

anthrotype:
> Longer-term I'd prefer to move away from AppVeyor and have #87 working instead.

true, but until it works and is free, reliable and fast I don't see an urgent reason to switch. The github windows runners have given me headaches sometimes (e.g. the temporary directory is a different drive than the one where the code was cloned to and os.path.relpath errors out when across mount points)

justvanrossum:
> github windows runners have given me headaches sometimes

Right, and this is another such headache...

PR #88, opened 2021-05-18
configure pytest to only look in tests/ by default 

anthrotype:
Attempt to fix https://github.com/harfbuzz/uharfbuzz/issues/86

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -64,6 +64,7 @@ commands =
     pip wheel --pre --no-deps --no-cache-dir --wheel-dir {distdir} --find-links {distdir} --no-binary {[tox]project_name} {[tox]project_name}
 
 [pytest]
+testpaths = tests/
 addopts =
     -v
     -r a


anthrotype:
yes, this seems to fix it

see that "Removing test/subset/__pycache__/"

https://ci.appveyor.com/project/fonttools/uharfbuzz/builds/39211316#L384

justvanrossum:
LGTM. If the CI completes succesfully, shall I merge and tag release v0.16.1?

anthrotype:
could it be that when we do `pytest`, it attempts to discover tests in all the subfolders including those that don't pertain to uharfbuzz itself? maybe we should exclude the harfbuzz dir or tell pytest to only look for tests inside the tests/ directory

justvanrossum:
Ahh, that could easily be it! I noticed that I had to run "pytest tests" because it was picking up some test-ish trash I havd lying around.

anthrotype:
yep, that's it

anthrotype:
now running `pytest` without arguments will default to `pytest tests/`

anthrotype:
> LGTM. If the CI completes succesfully, shall I merge and tag release v0.16.1?

yes please thanks!

Issue #89, opened 2021-06-01
Reading `Buffer.language` without setting it causes segmentation fault


kojiishi:
```
>>> import uharfbuzz as hb
>>> buffer = hb.Buffer()
>>> buffer.script
''
>>> buffer.language
zsh: segmentation fault  python3
```

This works when reading `language` after setting a value to it.

kojiishi:
Thank you Khaled!

PR #92, opened 2021-07-23
hb_language_to_string can return NULL

khaledhosny:
Return empty string in this case (for consistency with Buffer.script), but may be we should return None instead.

Fixes https://github.com/harfbuzz/uharfbuzz/issues/89


justvanrossum:
IMO `None` would be better, and I think it's ok to change Buffer.script to do the same.

justvanrossum (approved):

behdad:
Returning None is correct to me. Same for Script.

khaledhosny:
I did that. Invalid direction returns `invalid` though (from `hb_direction_to_string()`), I considered returning None here but then decided to keep it closer to what the C code does. If we ever do the API makeover, the functions should probably return enums not strings (`language` would still be a string, though).

PR #98, opened 2021-07-26
Drop AppVeyor CI builds

khaledhosny:
All three supported platforms are run in GitHub Actions now.


khaledhosny:
I don’t seem to have access to the uharfbuzz project on AppVeyor to remove it there.

khaledhosny:
> I don’t seem to have access to the uharfbuzz project on AppVeyor to remove it there.

I removed the webhooks from the repository settings though (removed also the Travis webhook we no longer need).

PR #99, opened 2021-08-01
Add Blob class

khaledhosny:
Allows creating blobs from data as will as file paths, and creating faces from them. Fixes https://github.com/harfbuzz/uharfbuzz/issues/52.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -325,22 +325,22 @@ cdef hb_blob_t* _reference_table_func(
 cdef class Face:
     cdef hb_face_t* _hb_face
     cdef object _reference_table_func
-    cdef object _blob
+    cdef Blob _blob
 
-    def __cinit__(self, bytes blob, int index=0):
-        cdef hb_blob_t* hb_blob
+    def __cinit__(self, blob: Union[Blob, bytes], int index=0):
         if blob is not None:
-            self._blob = blob
-            hb_blob = hb_blob_create(
-                blob, len(blob), HB_MEMORY_MODE_READONLY, NULL, NULL)
-            self._hb_face = hb_face_create(hb_blob, index)
-            hb_blob_destroy(hb_blob)
+            if not isinstance(blob, Blob):
+                self._blob = Blob(blob)
+            else:
+                self._blob = blob
+            self._hb_face = hb_face_create(self._blob._hb_blob, index)
         else:
             self._hb_face = NULL
 
     def __dealloc__(self):
         if self._hb_face is not NULL:
             hb_face_destroy(self._hb_face)
+        self._blob = None
 
     # DEPRECATED: use the normal constructor
     @classmethod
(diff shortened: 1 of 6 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -276,6 +277,30 @@ cdef class Buffer:
         hb_buffer_set_message_func(self._hb_buffer, msgcallback, <void*>callback, NULL)
 
 
+cdef class Blob:
+    cdef hb_blob_t* _hb_blob
+    cdef object _data
+
+    def __cinit__(self, bytes data):
```
khaledhosny, line 284: I'm wondering if it makes sense to allow `str` and `Path` here and use `from_file_path` in this case, or would that be too clever and confusing?

anthrotype: Hm no, i think it's better, less confusing to keep the two construction modes separate

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -102,8 +103,7 @@ cdef class Buffer:
         self._message_callback = None
 
     def __dealloc__(self):
-        if self._hb_buffer is not NULL:
-            hb_buffer_destroy(self._hb_buffer)
+        hb_buffer_destroy(self._hb_buffer)
```
anthrotype, line 106: I assume that the NULL check is redundant because it's already checked in hb_buffer_destroy, right?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -276,6 +276,30 @@ cdef class Buffer:
...
+        if data is not None:
+            self._data = data
+            self._hb_blob = hb_blob_create(
+                data, len(data), HB_MEMORY_MODE_READONLY, NULL, NULL)
+        else:
+            self._hb_blob = hb_blob_get_empty()
+
+    @classmethod
+    def from_file_path(cls, filename: Union[str, Path]):
+        cdef bytes packed = str(filename).encode()
```
anthrotype, line 293: I believe the correct way here would be to call os.fsencode() here, works on both str and PathLike objects and respects the filesystem encoding

anthrotype (approved):

src/uharfbuzz/_harfbuzz.pyx
khaledhosny: It is redundant because `hb_buffer_create()` never returns `NULL` as well as the other HarfBuzz create functions, they return a singleton in case of failure that are safe to pass to other HarfBuzz functions.

src/uharfbuzz/_harfbuzz.pyx
khaledhosny: OK.

src/uharfbuzz/_harfbuzz.pyx
khaledhosny: Done.

PR #101, opened 2021-08-19
Update to HarfBuzz 2.9.0

khaledhosny:
Also re-enable Uniscribe shaper, building it should be fixed now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,8 @@ requires = [
 build-backend = "setuptools.build_meta"
 
 [tool.cibuildwheel]
-# skip 2.7, 3.3, 3.4 and 3.5: for now, only build for >= 3.6 64-bit
-skip = ["cp27-*", "cp33-*", "cp34-*", "cp35-*", "*-manylinux_i686"]
+# only build for 64-bit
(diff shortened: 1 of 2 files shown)


khaledhosny:
I don’t know what to make of the build failure on Linux:
```
Building cp310-manylinux_x86_64 wheel
CPython 3.10 manylinux x86_64

Setting up build environment...
  
      + /opt/python/cp38-cp38/bin/python -c 'import sys, json, os; json.dump(os.environ.copy(), sys.stdout)'
      + which python
  cibuildwheel: python available on PATH doesn't match our installed instance. If you have modified PATH, ensure that you don't overwrite cibuildwheel's entry or insert python above it.
  Error: Process completed with exit code 1.
```

khaledhosny:
The breakage was apparently caused by the forced `manylinux1` image instead of using the default. I don’t know the exact cause, but [support for manylinux1 will end on January 1st, 2022](https://github.com/pypa/manylinux/issues/994), so I guess now is a good time to drop it.

khaledhosny:
Actually the above issue explicitly states that `manylinux1` will not support Python 3.10, so that is our root cause.

PR #110, opened 2022-02-15
Update to HarfBuzz 4.0.0

khaledhosny:
In preparation for the stabilization hb-draw API.

TODO:

- [x]  Review the Python API and align it with the new HarfBuzz API
- [x]  Deprecate old Python API if needed

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -1,5 +1,6 @@
 #cython: language_level=3
 import os
+import warnings
 from enum import IntEnum
 from .charfbuzz cimport *
 from libc.stdlib cimport free, malloc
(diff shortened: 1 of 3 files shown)


justvanrossum:
Blackrenderer uses the draw API, and may be a good test case. Let me know if/when you want me to try out this PR.

khaledhosny:
> Blackrenderer uses the draw API, and may be a good test case. Let me know if/when you want me to try out this PR.

It should be ready for testing now.

garretrieger:
Is this able to be merged now? We're planning on PR'ing the addition of the repacker API but it's blocked on updating to 4.0.0.

khaledhosny:
Merged, we can fix any issues before next release.

justvanrossum:
I just ran my blackrenderer tests with this version, and it works perfectly. 

I do see differences between the previous int-only output and the current float coordinates, but that's obviously to be expected. I'll just have to adapt my expected results in those few cases.

khaledhosny:
Excellent !

PR #113, opened 2022-03-15
Add hb_repack() for table packing

qxliu76:
add Harfbuzz repacker API


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,103 @@ cdef class DrawFuncs:
         self._close_path_func = func
         hb_draw_funcs_set_close_path_func(
             self._hb_drawfuncs, _close_path_func, <void*>user_data, NULL)
+
+cdef void free_allocated (hb_object_t* obj_list, unsigned int idx):
```
garretrieger, line 1061: rename idx to length.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,103 @@ cdef class DrawFuncs:
...
+    cdef unsigned int i
+    for i in range(idx):
+        if obj_list[i].real_links != NULL:
+            free(obj_list[i].real_links)
+        if obj_list[i].virtual_links != NULL:
+            free(obj_list[i].virtual_links)
+
+def hb_repack(subtables: List[bytes],
```
garretrieger, line 1069: Add a comment here documenting what the various parts of graphnodes are.

garretrieger (approved):

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,103 @@ cdef class DrawFuncs:
...
+        obj_list[i].tail = table_data + tail
+
+        node = graphnodes[i]
+        #real_links
+        real_link_num = len(node[0])
+        obj_list[i].num_real_links = real_link_num
+        if real_link_num == 0:
+            obj_list[i].real_links = NULL
+        else:
+            p = <hb_link_t*>calloc(real_link_num, sizeof(hb_link_t))
```
garretrieger, line 1112: This code for creating the link list (and the similar one for virtual links) could be extracted into a separate function to improve readability.

qxliu76: I tried to create a separate function but failed to build because:
This separate function needs to assign values to real_links member in hb_object_t struct with values coming from List[Tuple[int, int,int]]

for def function: i cannot pass in hb_object_t* type
for cdef function: I cannot pass in List[Tuple[int,int,int]]

qxliu76: Oh, I think I might be able to do this by defining a class which is a wrapper of hb_object_t*, I'll work on this

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+# - width: size of the offset (e.g 2 for offset16 and 4 for offset32)
+# - objidx: objidx is the index of the subtable in graph/tree generated by
+#           postorder traversal
+#
+# virtual_links is a list of virtual edges, they are not real offset links,
+# a virtual_link specifies an ordering constraint that harfbuzz packing must follow.
+# A virtual edge would have 0 offsetSize, 0 position, and a real objidx.
+# e.g, if Node A has a virtual edge with objidx b(corresponding Node is B),
+# then that means Node B is always packed after Node A in the final serialized order
+def hb_repack(subtables: List[bytes],
```
khaledhosny, line 1131: This comment should probably be a dicstring.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+        length += len(subtables[i])
+        idx_len_dict[i] = (prev_length, length)
+        prev_length = length
+
+    cdef bytes table_bytes = b''.join(subtables)
+    cdef char* table_data = table_bytes
+
+    cdef hb_object_t* obj_list = <hb_object_t*>calloc(num_nodes, sizeof(hb_object_t))
+    if obj_list == NULL:
+      return bytes(b'')
```
khaledhosny, line 1151: Shouldn’t this be a more obvious error than just returning an empty string? Same for the other cases below.

khaledhosny, line 1151: Shouldn’t this be a more obvious error than just returning an empty string?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+        if p == NULL:
+            free_allocated(obj_list, length)
+            # for virtual_links, need to free additional real_links which
+            # is created before virtual_links
+            if obj_list[length].real_links != NULL:
+                free(obj_list[length].real_links)
+            free(obj_list)
+
+            success[0] = 0
+            return NULL
```
khaledhosny, line 1093: Since the failure case returns NULL, is it necessary to track the success independently?

khaledhosny, line 1093: Since the failure case returns NULL, is it necessary to track the success independently?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+            self._hb_link[i].objidx = links[i][2]
+
+cdef hb_link_t* create_links(unsigned int link_num,
+                             hb_object_t* obj_list,
+                             unsigned int length,
+                             hb_bool_t* success):
+    success[0] = 1
+    cdef hb_link_t* p = NULL
+    if link_num == 0:
+        return NULL
```
khaledhosny, line 1081: Is it intentional that success is not set to zero here?

khaledhosny, line 1081: Is it intentional that success is not set to zero here?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+    cdef unsigned int i
+    for i in range(length):
+        if obj_list[i].real_links != NULL:
+            free(obj_list[i].real_links)
+        if obj_list[i].virtual_links != NULL:
+            free(obj_list[i].virtual_links)
+
+# The whole table is represented as a Graph
+# The input graphnodes is a flat list of subtables
+# each Node(subtable) is represented by a tuple of 2 lists: real_links and virtual_links
```
anthrotype, line 1107: consider defining a Link namedtuple instead of using un-named tuples, so you can do link.objidx instead of link[2], which is much more readable. You can still accept a plain tuple from the caller and then convert it to your namedtuple, or you could expose these objects in the public python api and the caller needs to import them and use them to construct the graph to pass on to hb_repacker

```python
from typing import NamedTuple

class Link(NamedTuple):
    width: int
    position: int
    objidx: int
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+# a virtual_link specifies an ordering constraint that harfbuzz packing must follow.
+# A virtual edge would have 0 offsetSize, 0 position, and a real objidx.
+# e.g, if Node A has a virtual edge with objidx b(corresponding Node is B),
+# then that means Node B is always packed after Node A in the final serialized order
+def hb_repack(subtables: List[bytes],
+              graphnodes: List[Tuple[List[Tuple[int, int, int]],
+                                     List[Tuple[int, int, int]]
+                                     ]]) -> bytes:
+    if len(subtables) != len(graphnodes):
+        return bytes(b'')
```
anthrotype, line 1136: `b''` is already a bytes literal, you don't need to cast to `bytes(b'')`

anthrotype: I don't like that errors are communicated to the caller with this implicit empty bytes string. I think you should raise python exceptions with error messages that make sense and let the caller handle them. In this case, it should be a ValueError probably because the contract of this method is that the two input lists must have the same length

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+        obj_list[i].tail = table_data + tail
+
+        node = graphnodes[i]
+        #real_links
+        real_link_num = len(node[0])
+        obj_list[i].num_real_links = real_link_num
+
+        p = create_links(real_link_num, obj_list, i, &success)
+        if success == 0:
+          return bytes(b'')
```
anthrotype, line 1176: when you fail to allocate some memory the right exception is MemoryError(), see elsewhere in this module

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,152 @@ cdef class DrawFuncs:
...
+            if obj_list[length].real_links != NULL:
+                free(obj_list[length].real_links)
+            free(obj_list)
+
+            success[0] = 0
+            return NULL
+        else:
+            return p
+
+cdef void free_allocated(hb_object_t* obj_list, unsigned int length):
```
anthrotype, line 1097: cython `cdef class` classes can have a `__dealloc__` method that is called when the object is garbage collected, and is used to free heap allocated memory, e.g. see

https://github.com/fonttools/skia-pathops/blob/0728fbdbf61b73a32207849026850a493b534cac/src/python/pathops/_pathops.pyx#L983-L996

the idea is you define a custom python type (cdef class) that can _own_ pointers to raw memory whose lifetime is tied to that of the python object itself, so you don't have to manually call malloc and free.
You can allocate the memory in the special `__cinit__` method (the default constructor for cdef class types which is called with the same parameters as the python `__init__` and where c stuff can be initialized) or define alternative class constructors (decorated with @staticmethod like the `_VerbArray.create` I linked to above) that return an instance of this cdef class, use it and then let the garbage collector call `__dealloc__` for you automatically when refcount is 0.

anthrotype (commented):
thank you Qunxin, I left some comments. I'm gonna be away the next couple weeks, but Khaled and Just can also review and cut releases.

src/uharfbuzz/_harfbuzz.pyx
qxliu76: yes, when NULL is returned, there're 2 cased:
num of links = 0, success =1
num of links > 0 , success =0, means memory allocation failed. 
So we need another boolean flag here to indicate whether memory allocating succeeds.

qxliu76: close duplicate comment

src/uharfbuzz/_harfbuzz.pyx
qxliu76: yes, when NULL is returned, there're 2 cased:
num of links = 0, success =1
num of links > 0 , success =0, means memory allocation failed.
So we need another boolean flag here to indicate whether memory allocating succeeds.

qxliu76: close duplicate comment

src/uharfbuzz/_harfbuzz.pyx
qxliu76: typing.NamedTuple doesn't seem to work with python, see also:
https://github.com/harfbuzz/uharfbuzz/blob/main/src/uharfbuzz/_harfbuzz.pyx#L374

qxliu76:
I have addressed all review comments(refactored code in python style), could you help review again? @khaledhosny

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,146 @@ cdef class DrawFuncs:
...
+            #real_links
+            p = obj_list.create_links(i, len(node[0]), 1)
+            if p != NULL:
+                obj_list.update_links(i, 1, node[0])
+            #virtual_links
+            p = obj_list.create_links(i, len(node[1]), 0)
+            if p != NULL:
+                obj_list.update_links(i, 0, node[1])
+    except MemoryError:
+        raise MemoryError()
```
khaledhosny, line 1189: This is re-raising the same exception, which makes the try except block redundant and can be done without.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+    cdef update_obj_length(self,
+                           unsigned int idx,
+                           char* head,
+                           char* tail):
+        self._hb_obj_list[idx].head = head
+        self._hb_obj_list[idx].tail = tail
+       
+    cdef hb_link_t * create_links(self, unsigned int idx,
+                                  unsigned int link_num,
+                                  int is_real_link):
```
khaledhosny, line 1089: Shouldn’t this be `bint` instead of `int`?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+        if link_num == 0:
+            return NULL
+        cdef hb_link_t* p = <hb_link_t*>calloc(link_num, sizeof(hb_link_t))
+        if p == NULL:
+            raise MemoryError()
+        else:
```
khaledhosny, line 1095: Since the other branch raises, no need for `else` here.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+            else:
+                self._hb_obj_list[idx].num_virtual_links = link_num
+                self._hb_obj_list[idx].virtual_links = p
+            return p
+
+    def update_links(self, unsigned int idx, int is_real_link,
+                     links: List[Tuple[int, int, int]]):
+        cdef unsigned int num_links = len(links)
+        cdef hb_link_t* l = NULL
+        if is_real_link == 1:
```
khaledhosny, line 1108:
```suggestion
        if is_real_link:
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+    cdef hb_link_t * create_links(self, unsigned int idx,
+                                  unsigned int link_num,
+                                  int is_real_link):
+        if link_num == 0:
+            return NULL
+        cdef hb_link_t* p = <hb_link_t*>calloc(link_num, sizeof(hb_link_t))
+        if p == NULL:
+            raise MemoryError()
+        else:
+            if is_real_link == 1:
```
khaledhosny, line 1096:
```suggestion
            if is_real_link:
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+                self._hb_obj_list[idx].num_real_links = link_num
+                self._hb_obj_list[idx].real_links = p
+            else:
+                self._hb_obj_list[idx].num_virtual_links = link_num
+                self._hb_obj_list[idx].virtual_links = p
+            return p
+
+    def update_links(self, unsigned int idx, int is_real_link,
```
khaledhosny, line 1104:
```suggestion
    def update_links(self, unsigned int idx, bint is_real_link,
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+            free(self._hb_obj_list)
+
+    cdef update_obj_length(self,
+                           unsigned int idx,
+                           char* head,
+                           char* tail):
+        self._hb_obj_list[idx].head = head
+        self._hb_obj_list[idx].tail = tail
+       
+    cdef hb_link_t * create_links(self, unsigned int idx,
```
khaledhosny, line 1087:
```suggestion
    cdef hb_link_t* create_links(self, unsigned int idx,
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+        hb_link_t* p = NULL
+        HBObject obj_list = HBObject(num_nodes)
+    for i in range(num_nodes):
+        tail += len(subtables[i])
+        obj_list.update_obj_length(i, table_data + head, table_data + tail)
+        head = tail
+        
+        node = graphnodes[i]
+        #real_links
+        p = obj_list.create_links(i, len(node[0]), 1)
```
khaledhosny, line 1180:
```suggestion
        p = obj_list.create_links(i, len(node[0]), True)
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+        if p != NULL:
+            obj_list.update_links(i, 1, node[0])
```
khaledhosny, line 1182:
```suggestion
            obj_list.update_links(i, True, node[0])
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+        unsigned int head = 0, tail = 0
+        hb_link_t* p = NULL
+        HBObject obj_list = HBObject(num_nodes)
+    for i in range(num_nodes):
+        tail += len(subtables[i])
+        obj_list.update_obj_length(i, table_data + head, table_data + tail)
+        head = tail
+        
+        node = graphnodes[i]
+        #real_links
```
khaledhosny, line 1179:
```suggestion
        # real links
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+        p = obj_list.create_links(i, len(node[0]), 1)
+        if p != NULL:
+            obj_list.update_links(i, 1, node[0])
+        #virtual_links
```
khaledhosny, line 1183:
```suggestion
        # virtual inks
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1057,3 +1057,143 @@ cdef class DrawFuncs:
...
+        p = obj_list.create_links(i, len(node[1]), 0)
+        if p != NULL:
+            obj_list.update_links(i, 0, node[1])
+
+    cdef hb_blob_t*  packed_blob = hb_subset_repack_or_fail(obj_list._hb_obj_list, num_nodes)
```
khaledhosny, line 1188:
```suggestion
    cdef hb_blob_t* packed_blob = hb_subset_repack_or_fail(obj_list._hb_obj_list, num_nodes)
```

src/uharfbuzz/_harfbuzz.pyx
khaledhosny, line 1104: Any particular reason this uses `def` and `create_links` uses `cdef`?

src/uharfbuzz/_harfbuzz.pyx
khaledhosny: Any reason collections.namedtuple can’t be used here then?

qxliu76: collections.namedtuple doesn't support type hints.
Besides, it can not be defined as a subclass,  so I need to define it in both fonttools and cython.

khaledhosny (approved):

anthrotype:
a belated LGTM! thank you Qunxin!! 👍

Issue #116, opened 2022-03-28
`Blob.from_file_path` does not bail on error


madig:
The following snippet will not raise an error:

```py
import uharfbuzz as hb
hb.Blob.from_file_path("asdasgdjh")
```

It should probably use https://harfbuzz.github.io/harfbuzz-hb-blob.html#hb-blob-create-from-file-or-fail under the hood?

anthrotype:
Yes it should

khaledhosny:
The C API  does not bail on error by design, I think there is an issue for this on HarfBuzz bug tracker and ways to detect failure.

behdad:
In C the way is to check for equality to the empty blob. But in Python I think using the create_from_file_or_fail and throwing exception is fine...

khaledhosny:
I forgot about the or_fail variant.

PR #118, opened 2022-04-01
Raise on failure in Blob.from_file_path()

khaledhosny:
By using `hb_blob_create_from_file_or_fail()` and raising when it returns `NULL`.

Fixes https://github.com/harfbuzz/uharfbuzz/issues/116

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -38,6 +38,11 @@ def version_string() -> str:
     cdef bytes packed = cstr
     return packed.decode()
 
+
+class HarfBuzzError(Exception):
+    pass
+
+
 cdef class GlyphInfo:
     cdef hb_glyph_info_t _hb_glyph_info
     # could maybe store Buffer to prevent GC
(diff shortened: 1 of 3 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -293,8 +293,11 @@ cdef class Blob:
     @classmethod
     def from_file_path(cls, filename: Union[str, Path]):
         cdef bytes packed = os.fsencode(filename)
+        cdef hb_blob_t* blob = hb_blob_create_from_file_or_fail(<char*>packed)
+        if blob == NULL:
+            raise FileNotFoundError()
```
khaledhosny, line 298: I’m not sure if this is the right exception since it can also fail due to memory allocation or other issues, but I’m not sure what generic exception is better here or if I should make a new one.

madig: Maybe a new `UHarfBuzzException`? I think it's appropriate to have library errors.

madig (approved):

madig:
Out of curiosity, does HB provide any more detail on why some operations fail?

khaledhosny:
No, it doesn't.

PR #125, opened 2022-07-01
Add convenience Font.from_file_path

madig:
Because why not.

Can't easily test this locally because I don't have Cython handy. I think tox.ini is missing an `isolated_build = true`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 project_name = uharfbuzz
 envlist = py3,py3-cov,coverage
 skip_missing_interpreters = true
+isolated_build = true
 
 [testenv]
 skip_install =
--- a/README.md
+++ b/README.md
@@ -17,9 +17,7 @@ import uharfbuzz as hb
 fontfile = sys.argv[1]
 text = sys.argv[2]
 
-blob = hb.Blob.from_file_path(fontfile)
-face = hb.Face(blob)
-font = hb.Font(face)
+font = hb.Font.from_file_path(fontfile)
 
 buf = hb.Buffer()
 buf.add_str(text)
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -408,6 +408,23 @@ cdef class Font:
         cdef Font inst = cls(face)
         return inst
 
+    @classmethod
+    def from_file_path(cls, filename: Union[str, Path], face_index: int = 0):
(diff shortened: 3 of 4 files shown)


tox.ini
```diff
@@ -2,6 +2,7 @@
 project_name = uharfbuzz
 envlist = py3,py3-cov,coverage
 skip_missing_interpreters = true
+isolated_build = true
```
anthrotype, line 5: I think I didn't add that on purpose back in the days. I don't want you to build the package itself because I am running setup.py develop myself so that some environment variable are set and enable some coverage tracing.

anthrotype: I meant "tox", not "you" - misstyping from mobile phone

anthrotype: Try to see if you can disable isolated build at least when tox env contains the "cov" particle (e.g. py310-cov):


isolated_build =
    cov: false
    !cov: true

I'm not sure if that is going to work but that was my intention

anthrotype:
As for the new API itself, I'm not sure. uharfbuzz is quite low level and tries to mimic the harfbuzz API quite closely.
But I suppose it doesn't hurt either.
I'll leave it to @justvanrossum or @khaledhosny to decide

tox.ini
madig: That doesn't seem to work. I wonder if coverage etsting can be done more "cleanly"...

khaledhosny:
I like to expose the HarfBuzz C API faithfully as it allows for writing performance code. I don’t mind higher level API, though, if it allows people to write badly performing code 🤷🏾

tox.ini
anthrotype: Then just disable it for now

anthrotype: The isolated build I mean

madig:
What makes the method inefficient, except if you want to instantiate multiple Fonts from the same Blob or Face? Where would you do that?

khaledhosny:
In font collections (TTC, dfont) you can create multiple faces from the same blob. Fonts you can either create and reuse one font per size/variation settings etc, or keep changing the same font possibly invalidating internal HarfBuzz caches everytime the font changes.

madig:
Hm. So it helps to instantiate multiple Fonts explicitly from the same Face when you want to test VFs across the design space?

khaledhosny:
Not only for testing. Supposed to want to use bold and regular in the same document and it is a variable font, you create two fonts with different variation settings, or you create one font and keep changing variation settings each time you want to switch weights.

PR #126, opened 2022-07-02
Start using new setuptools config stuff

madig:
The packaging ecosystem is converging to using pyproject.toml for everything: https://setuptools.pypa.io/en/stable/userguide/pyproject_config.html

Here, I move stuff to using it, because I can.

The tox wheel env can go maybe, one can use [`build`](https://pypi.org/project/build/) instead.

Needs more testing.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,11 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         submodules: recursive
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: "3.x"
     - name: Install dependencies
@@ -43,17 +43,17 @@ jobs:
     needs: [build]
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
       with:
         submodules: recursive
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: "3.x"
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install --upgrade setuptools cython wheel twine
+        pip install --upgrade setuptools cython wheel twine build
     - name: Download artifacts from build jobs
       uses: actions/download-artifact@v2
       with:
@@ -108,5 +108,5 @@ jobs:
         else
           echo "DEBUG: This is a final release"
         fi
-        python setup.py sdist
+        python -m build --sdist
         twine upload dist/*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,11 @@ jobs:
       with:
         python-version: "3.x"
     - name: Install dependencies
-      run: pip install cibuildwheel
+      run: pip install cibuildwheel twine
     - name: Build Wheels
-      run: python -m cibuildwheel --output-dir wheelhouse
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
+        twine check wheelhouse/*
     - uses: actions/upload-artifact@v2
       with:
         name: uharfbuzz-${{ matrix.os }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,9 +30,9 @@ jobs:
     - name: Install dependencies
       run: pip install cibuildwheel twine
     - name: Build Wheels
-      run: |
-        python -m cibuildwheel --output-dir wheelhouse
-        twine check wheelhouse/*
+      run: python -m cibuildwheel --output-dir wheelhouse
+    - name: Check Wheels
+      run: twine check wheelhouse/*
     - uses: actions/upload-artifact@v2
       with:
         name: uharfbuzz-${{ matrix.os }}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "uharfbuzz"
-description = "Streamlined Cython bindings for the harfbuzz shaping engine"
+description = "Streamlined Cython bindings for the HarfBuzz shaping engine"
 readme = "README.md"
 license = {file = "LICENSE"}
 requires-python = ">=3.5"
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -1,12 +1,28 @@
...
     "wheel",
-    "setuptools_scm >= 2.1",
+    "setuptools_scm[toml] >= 6.2",
     "cython >= 0.28.1",
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "uharfbuzz"
+description = "Streamlined Cython bindings for the harfbuzz shaping engine"
```
khaledhosny, line 12: May be capitalize HarfBuzz while at it.

pyproject.toml
```diff
@@ -1,12 +1,28 @@
...
+readme = "README.md"
+license = {file = "LICENSE"}
+requires-python = ">=3.5"
+authors = [
+    { name = "Adrien Tétar", email = "adri-from-59@hotmail.fr" },
```
khaledhosny, line 17: We need to update this as well.

pyproject.toml
```diff
@@ -1,12 +1,28 @@
...
+]
+dynamic = [
+    "version",
+]
+
+[project.urls]
+Homepage = "https://github.com/trufont/uharfbuzz"
```
khaledhosny, line 24: And this.

anthrotype:
> because I can.

honestly, I don't like change just because of change.. You could elaborate a bit on why this is needed, but whatever. As long as I can continue to run `setup.py build_ext --inplace` _outside_ of an isolated build, I'm fine with this.
Yeah, it's cool that python packaging is converging towards pyproject.toml, but editable installs are still not supported in a pure setup.py-less world AFAIK, and they are too convenient to drop so we better maintiain support for those if we can.

> The tox wheel env can go maybe

yeah, that's unused.

> one can use [build](https://pypi.org/project/build/) instead.

sure, to build the sdist at least, For the wheels we must use cibuildwheel

madig:
> editable installs are still not supported in a pure setup.py-less world AFAIK

Weird. This seems to be setuptools-specific (works with hatch AFAICS) but the docs mention it should work with up to date pips: https://setuptools.pypa.io/en/stable/userguide/quickstart.html#development-mode. Doing a quick test locally gives me an error message though, even with up to date setuptools and pip 🤷 Since a setup.py is needed anyway for extension building, I don't see a problem though.

> As long as I can continue to run setup.py build_ext --inplace outside of an isolated build

Seems to work.

pyproject.toml
madig: Who should be put in and in what order?

pyproject.toml
```diff
@@ -1,12 +1,28 @@
...
     "cython >= 0.28.1",
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "uharfbuzz"
+description = "Streamlined Cython bindings for the HarfBuzz shaping engine"
+readme = "README.md"
+license = {file = "LICENSE"}
+requires-python = ">=3.5"
```
madig, line 15: Should this be set to `>= 3.6`?

anthrotype: i guess, since we no longer build wheels for 3.5 so we are not testing it agaist that. It's EOL so we may as well explictly drop support

madig: Oh, actually. setupttols 62.5 requires >= 3.7. Don't mind if I do :)

anthrotype: now.. these are the kind of breaking changes for change sake that I really dislike.

anthrotype: hold on! python3.6 also is EOL. I got confused, sorry. 😅 
yes, that's fine then

madig:
The wheels contain Cython sources 🤔

anthrotype:
> The wheels contain Cython sources

so what?

madig:
Also, building the binary wheels floods my screen with (MSVC) compiler warnings from the HarfBuzz source 🤨

madig:
> so what?

The pypi wheels don't, but yes.

anthrotype:
> The pypi wheels don't, but yes.

i'm confused. What are you trying to say? I don't see a problem if the .pyx sources are included in the wheel alongside the .py files, as long as the compiled .so extensions are there.

pyproject.toml
khaledhosny: I don’t know, HarfBuzz Team or something. Do we even need to explicitly set an author?

madig:
> I don't see a problem if the .pyx sources are included in the wheel

Agree.

pyproject.toml
madig: Simply removing it seems to at least build and install in tox. I can remove it if everyone agrees?

anthrotype: I think some author should be there. FWIW I don't see a problem keeping like it was, or if you want to add The HarfBuzz Team that's fine too.
In general, you can run `twine check` on the wheel to see if it complains about any missing metadata.

madig:
Oof, building wheels in parallel did the opposite of help:

* serial: https://github.com/harfbuzz/uharfbuzz/actions/runs/2601563856
* parallel: https://github.com/harfbuzz/uharfbuzz/actions/runs/2606579967

anthrotype:
interesting, was worth trying though

pyproject.toml
madig: Wouldn't you have to define that somewhere? I'll just add khaledhosny, justvanrossum and anthrotype in laphabetical order for now.

.github/workflows/ci.yml
```diff
@@ -28,9 +28,11 @@ jobs:
       with:
         python-version: "3.x"
     - name: Install dependencies
-      run: pip install cibuildwheel
+      run: pip install cibuildwheel twine
     - name: Build Wheels
-      run: python -m cibuildwheel --output-dir wheelhouse
+      run: |
+        python -m cibuildwheel --output-dir wheelhouse
+        twine check wheelhouse/*
```
anthrotype, line 35: how about running that in a subsequent separate step, separate from the `Build Wheels` one?

anthrotype:
> The pypi wheels don't [contain Cython sources]...

interesting. I just tried to download and `wheel unpack` a wheel from https://pypi.org/project/uharfbuzz/#files and I can see that it does _not_ contain any `.pyx` cython source files, only `.so` compiled extension module.
Whereas, the ones from your GH Action's artifacts contain those (as well as .cpp sources compiled from those).

What exactly changed to produce this is unclear, and I think deserves further investigation.

madig:
> You could elaborate a bit on why this is needed

It's not. My idea is that a little update here and there could prevent bigger headaches down the road. I fixed a couple of packing warnings along the way (e.g. 2f7e2574397a140f3bd18b4ed27817c6baf6fa9e and 55932e7b902fe70aa63280d86fc121715531e6b2).

PR #130, opened 2022-07-26
Allow passing capsules for draw callbacks

khaledhosny:
This allow defining the callbacks in C modules and passing the pointers as Python objects.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -6,6 +6,7 @@ from .charfbuzz cimport *
 from libc.stdlib cimport free, malloc, calloc
 from libc.string cimport const_char
 from collections import namedtuple
+from cpython.pycapsule cimport PyCapsule_GetPointer, PyCapsule_IsValid
 from typing import Callable, Dict, List, Sequence, Tuple, Union
 from pathlib import Path
 
@@ -1033,9 +1034,15 @@ cdef class DrawFuncs:
                                      object,  # user_data
                                  ], None],
                                  user_data: object = None) -> None:
-        self._move_to_func = func
+        cdef hb_draw_move_to_func_t func_p
+        if PyCapsule_IsValid(func, NULL):
+            self._move_to_func = None
+            func_p = <hb_draw_move_to_func_t>PyCapsule_GetPointer(func, NULL)
+        else:
+            self._move_to_func = func
+            func_p = _move_to_func
         hb_draw_funcs_set_move_to_func(
-            self._hb_drawfuncs, _move_to_func, <void*>user_data, NULL)
+            self._hb_drawfuncs, func_p, <void*>user_data, NULL)
 
     def set_line_to_func(self,
                          func: Callable[[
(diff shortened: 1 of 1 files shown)


behdad:
This is neat.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1033,9 +1034,14 @@ cdef class DrawFuncs:
                                      object,  # user_data
                                  ], None],
                                  user_data: object = None) -> None:
-        self._move_to_func = func
+        cdef hb_draw_move_to_func_t func_p
+        if PyCapsule_IsValid(func, NULL):
+            func_p = <hb_draw_move_to_func_t>PyCapsule_GetPointer(func, NULL)
```
justvanrossum, line 1039: For completeness sake, this should probably set `self._move_to_func` to None, just in case this call is overriding a previous call.

justvanrossum:
Should `user_data` also be allowed to be a capsule?

khaledhosny:
> Should `user_data` also be allowed to be a capsule?

I wondered about that too, but it seems to work without a change. I guess because `user_data` is accessed from the callback functions, so that is the part the needs to deal with whether it is a capsule or a another Python object.

justvanrossum:
I'm just worried that the callback code will have to deal with unpacking the capsule. If it requires a Python C API call that would be unfortunate.

khaledhosny:
Good point.

justvanrossum:
Can capsules interoperate with ctypes pointers? It's a bit unclear to me how compatible the two are.

khaledhosny:
> Can capsules interoperate with ctypes pointers? It's a bit unclear to me how compatible the two are.

I think they should work, PyCapsule is essentially wrapping an integer in a Python object. The problem, I think, will be that there is no Python API for dealing with PyCapsule’s, only a C-API.

anthrotype:
I think ctypes.pythonapi exposes the Python API, including PyCapsule, not sure if that's what you're after.
I do think cython is neater than ctypes in general.

anthrotype:
found this on https://stackoverflow.com/a/65057267

anthrotype (approved):

justvanrossum (approved):
LGTM, and works well in FontGoggles. Some tests would be nice, but we can do that later.

khaledhosny:
Testing would be good indeed, but I was hesitant to look into it because it will need some compiled native code and I have no idea how to pull that for tsets.

justvanrossum:
Could it be faked with some cdefs in the main .pyx source?

khaledhosny:
I don’t know, I’m apparently not a Cython expert. @anthrotype probably has some ideas.

anthrotype:
maybe you could define some dummy move_to, line_to etc. methods inside the main .pyx module (or a new extension module that's only for testing but that complicates the setup), then from inside the pure-python test module you use ctypes to `ctypes.cdll.LoadLibrary` the uharfbuzz extension module itself in the same way FontGoggles loads its own objc dylib (an extension module is like a shared library, I think) and then use ctypes to make the capsules and pass them on to uharfbuzz DrawFuncs..

anthrotype:
I tried just for fun to load the uharfbuzz extension via ctypes, looks like it works

```python
In [1]: import ctypes
In [2]: import uharfbuzz._harfbuzz
In [3]: uharfbuzz._harfbuzz.__file__
Out[3]: '/Users/clupo/Github/nanoemoji/.venv/lib/python3.10/site-packages/uharfbuzz/_harfbuzz.cpython-310-darwin.so'
In [4]: lib = ctypes.cdll.LoadLibrary(uharfbuzz._harfbuzz.__file__)
In [5]: hb_version_string = lib.hb_version_string
In [6]: hb_version_string.restype = ctypes.c_char_p
In [7]: hb_version_string.argtypes = ()
In [8]: hb_version_string()
Out[8]: b'5.0.0'
```

I think the dummy move_to, line_to, etc. for testing should be defined as `cdef` methods (so they are only visible from C)

khaledhosny:
Thanks, I’ll give this a try.

PR #131, opened 2022-07-29
Test passing PyCapsule to DrawFuncs callbacks

khaledhosny:


behdad:
Wow. This is some magic @justvanrossum would love!

khaledhosny:
> Wow. This is some magic @justvanrossum would love!

Already in latest FontGoggles https://github.com/justvanrossum/fontgoggles/pull/269

justvanrossum:
Deploy first, test later :)

khaledhosny:
This is failing on Windows, and with PyPy on Linux. No idea if any of this is expected.

khaledhosny:
> Deploy first, test later :)

FontGoogles **is** the test, this PR is just the paper work 😄

khaledhosny:
The failures are because ctypes can’t find the C callback functions. There are probably some magic incarnation to export the symbols in all configurations, but I have no idea what is it. May be we should skip the test if the symbols can’t be found? Or xfail on the known to fail combinations so that ones that pass now don’t silently fail later?

justvanrossum:
Could it be `__declspec`? https://stackoverflow.com/questions/2284610/what-is-declspec-and-when-do-i-need-to-use-it

khaledhosny:
> Could it be `__declspec`? https://stackoverflow.com/questions/2284610/what-is-declspec-and-when-do-i-need-to-use-it

I tried to use `HB_EXTERN` which supposedly handle this, but I now tried setting it directly.

behdad:
`HB_DLL_EXPORT` is internal to harfbuzz and not checked by the external headers.

khaledhosny:
> `HB_DLL_EXPORT` is internal to harfbuzz and not checked by the external headers.

I see. Setting `__declspec` directly seems to be working.

khaledhosny:
Test now xfails on PyPy, because its ctypes module has no pythonapi attribute and I couldn’t find a replacement.

justvanrossum (approved):

PR #132, opened 2022-07-29
Update hb_subset_repack_or_fail to new method signature that takes table tag.

garretrieger:


khaledhosny:
The test failure seems to be caused by changes in the message callbacks, please update the test expectations.

anthrotype:
I guess it's a bit too late, but I think it might have been nicer or more pythonic to add an optional `table_tag=""` keyword argument to the existing `repack` method, instead of defining/exporting an additional `repack_with_tag` method in the public uharfbuzz API. I guess you did that to more closely match Harfbuzz's own C API.

PR #134, opened 2022-07-30
[draw API] Avoid some overhead in font.draw_glyph_with_pen()

justvanrossum:
To draw with a pen, don't go via the Python DrawFuncs layer, but use C callbacks directly.

When using a RecordingPen, this is about 25-30% faster.

I mostly wrote this to learn more about the draw API. It adds some lower level code for relatively little gain. I do think it's an improvement overall, but won't complain if it gets rejected.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -546,24 +546,61 @@ cdef class Font:
         return packed.decode()
 
     def draw_glyph_with_pen(self, gid: int, pen):
-        funcs = DrawFuncs()
-        def move_to(x,y,c):
-            c.moveTo((x,y))
-        def line_to(x,y,c):
-            c.lineTo((x,y))
-        def cubic_to(c1x,c1y,c2x,c2y,x,y,c):
-            c.curveTo((c1x, c1y), (c2x, c2y), (x,y))
-        def quadratic_to(c1x,c1y,x,y,c):
-            c.qCurveTo((c1x, c1y), (x,y))
-        def close_path(c):
-            c.closePath()
+        cdef hb_draw_funcs_t* drawfuncs = hb_draw_funcs_create()
+        try:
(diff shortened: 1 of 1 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -546,24 +546,61 @@ cdef class Font:
...
-        funcs.set_close_path_func(close_path, pen)
-        funcs.get_glyph_shape(self, gid)
+        cdef hb_draw_funcs_t* drawfuncs = hb_draw_funcs_create()
+        try:
+            hb_draw_funcs_set_move_to_func(drawfuncs, _pen_move_to_func, NULL, NULL)
+            hb_draw_funcs_set_line_to_func(drawfuncs, _pen_line_to_func, NULL, NULL)
+            hb_draw_funcs_set_cubic_to_func(drawfuncs, _pen_cubic_to_func, NULL, NULL)
+            hb_draw_funcs_set_quadratic_to_func(drawfuncs, _pen_quadratic_to_func, NULL, NULL)
+            hb_draw_funcs_set_close_path_func(drawfuncs, _pen_close_path_func, NULL, NULL)
+            hb_font_get_glyph_shape(self._hb_font, gid, drawfuncs, <void*>pen)
```
khaledhosny, line 556: While at it, can we set the draw functions once and reuse them?

justvanrossum: We could, but since this it at the C level, I don't think this will help much. It's one allocation and 5 C calls.

khaledhosny: One `calloc`/`free` call per glyph, these things add up (I’m kidding, that is probably so tiny to be measurable anyway).

justvanrossum: I bet a no-op Python moveTo call is at least a magnitude more expensive :)

khaledhosny (approved):
Thanks!  25-30% faster is not nothing.

PR #135, opened 2022-07-31
[draw API] Get another 10-15% speedup by pre-fetching the pen instance methods

justvanrossum:
Pre-fetching the pen instance methods and passing them via `user_data` speeds up pen drawing by another 10-15%.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -547,6 +547,8 @@ cdef class Font:
 
     def draw_glyph_with_pen(self, gid: int, pen):
         cdef hb_draw_funcs_t* drawfuncs = hb_draw_funcs_create()
+        # keep local references to the bound pen methods to avoid them being gc'ed
+        # before they're even called
         pen_moveTo = pen.moveTo
         pen_lineTo = pen.lineTo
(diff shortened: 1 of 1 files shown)


behdad:
I'm glad you are finding the user_data useful!

justvanrossum:
> I'm glad you are finding the user_data useful!

Totally! It appears to me this is exactly what `user_data` is good at: passing method-specific implementations. `DrawFuncs` objects should do the same for the pure Python callbacks.

behdad:
> Totally! It appears to me this is exactly what `user_data` is good at: passing method-specific implementations

Correct. That's what they are for.

behdad:
Typically the user_data is bound to the method to call for that particular call, whereas the draw_data is bound to the instance.

behdad:
I've got to say what you and Khaled are doing with this API is all magic to me.

justvanrossum:
Right. In this case `user_data` is both, as I'm using it to pass bound methods.

khaledhosny:
> I've got to say what you and Khaled are doing with this API is all magic to me.

A well designed API allows its users to do things not imagined by its designers 😉  (I think there was a saying somehow like that, or may be I’m making things up)

PR #139, opened 2022-10-31
Add hb_font.create_sub_font(font)

behdad:
Untested.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -409,6 +409,12 @@ cdef class Font:
         cdef Font inst = cls(face)
         return inst
 
+    def create_sub_font(self):
+        cdef Font inst = self.__class__.__new__(self.__class__)
+        inst._hb_font = hb_font_create_sub_font (self._hb_font)
+        inst._face = self._face
+        return inst
+
     @property
     def face(self):
         return self._face
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -395,9 +395,19 @@ cdef class Font:
     cdef Face _face
     cdef FontFuncs _ffuncs
 
-    def __cinit__(self, Face face):
+    def __init__(self, face_or_font):
+        if isinstance(face_or_font, Font):
+            self.__create_sub_font(face_or_font)
+            return
+        self.__create(face_or_font)
+
+    cdef __create(self, Face face):
         self._hb_font = hb_font_create(face._hb_face)
         self._face = face
+
+    cdef __create_sub_font(self, Font font):
+        self._hb_font = hb_font_create_sub_font (font._hb_font)
+        self._face = font._face
 
     def __dealloc__(self):
         hb_font_destroy(self._hb_font)
@@ -407,12 +417,6 @@ cdef class Font:
     @classmethod
     def create(cls, face: Face):
         cdef Font inst = cls(face)
-        return inst
-
-    def create_sub_font(self):
-        cdef Font inst = self.__class__.__new__(self.__class__)
-        inst._hb_font = hb_font_create_sub_font (self._hb_font)
-        inst._face = self._face
         return inst
 
     @property
--- a/src/uharfbuzz/charfbuzz.pxd
+++ b/src/uharfbuzz/charfbuzz.pxd
@@ -271,6 +271,7 @@ cdef extern from "hb.h":
     void hb_font_funcs_destroy(hb_font_funcs_t* ffuncs)
 
     hb_font_t* hb_font_create(hb_face_t* face)
+    hb_font_t* hb_font_create_sub_font(hb_font_t* parent)
     void hb_font_set_funcs(
         hb_font_t* font,
         hb_font_funcs_t* klass,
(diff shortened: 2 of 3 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -409,6 +409,11 @@ cdef class Font:
         cdef Font inst = cls(face)
         return inst
 
+    def create_sub_font(self, parent: Font):
+        cdef Font inst = self.__class__()
```
anthrotype, line 413: I belive to instantiate a uharfbuzz.Font also you need to pass in a uharfbuzz.Face as well, see `__cinit__` above which takes a positional `face` parameter. Maybe just setting that to `parent.face` should work for your use case.

Also this new `create_sub_font` method would better be decorated with `@classmethod` and take a `cls` as first parameter rathen than `self` (for instance bound methods), since the way you wrote it, it behaves like an alternative constructor.

Actually, you could turn this into an instance method if you like, and then simply use it as `font.create_sub_font()` instead of `Font.create_sub_font(font)`. E.g.:

```cython
    def create_sub_font(self):
        cdef Font inst = self.__class__(self.face)
        inst._hb_font = hb_font_create_sub_font(self._hb_font)
        return inst
```

anthrotype: hm my snippet above may leak because I haven't called hb_font_destroy on the hb_font created inside `__cinit__` when I did `self.__class__(self.face)`...

Better to avoid calling `__cinit__` altogether and use `__new__` like so (untested):

```cython
    def create_sub_font(self):
        cdef Font inst = self.__class__.__new__(self.__class__)
        inst._hb_font = hb_font_create_sub_font(self._hb_font)
        inst._face = self._face
        return inst
```

behdad: Instance method was my intention indeed. I'm undecided as to which one makes most sense.

anthrotype: +1 for instance method, even though it's not identical to the HB C-api

behdad:
PTAL.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -409,6 +409,11 @@ cdef class Font:
         cdef Font inst = cls(face)
         return inst
 
+    def create_sub_font(self, parent: Font):
+        cdef Font inst = self.__class__(self.face)
```
anthrotype, line 413: i think this leads to a memory leak because we allocate a new inst._hb_font from the same self.face (self.__class__() uses the default constructor, hence it calls `__cinit__`), then immediately afterwards we are discarding that and setting inst._hb_font without freeing the memory (as we it's done in `__dealloc__`).

I think the current approach for alternate constructors is to skip `__cinit__` altogether and use `__new__` instead to get an instance of the class with fields still not initialized and initialize those explicitly like a constructor would do. 

```cython
    def create_sub_font(self):
        cdef Font inst = self.__class__.__new__(self.__class__)
        inst._hb_font = hb_font_create_sub_font(self._hb_font)
        inst._face = self._face
        return inst
```

behdad: Done.

anthrotype:
it would be good to confirm that it actually works as intended with a little test in https://github.com/harfbuzz/uharfbuzz/blob/main/tests/test_uharfbuzz.py

even as simple as loading a font, creating the sub font from it, then asserting that the returned value is actually a Font and that its id() is different (font2 is not font) even though the Face the two fonts refer to is the same (e.g. font2.face == font.face)

behdad:
> it would be good to confirm that it actually works as intended with a little test in https://github.com/harfbuzz/uharfbuzz/blob/main/tests/test_uharfbuzz.py
> 
> even as simple as loading a font, creating the sub font from it, then asserting that the returned value is actually a Font and that its id() is different (font2 is not font) even though the Face the two fonts refer to is the same (e.g. font2.face == font.face)

Thanks. Done. It fails. I don't know pyx enough to know what I'm doing wrong...

behdad:
Looks like calling `__new__` still calls `__cinit__` after. Is that expected?

behdad:
I'm also getting a weird error that I don't see on the CI:
```
>       font2 = font.create_sub_font()
E       TypeError: Font.create_sub_font() takes exactly one argument (0 given)
```

anthrotype:
I'll take a look tomorrow

behdad:
I managed to move it to constructor and have tests pass.

anthrotype (approved):
LGTM

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -395,10 +395,20 @@ cdef class Font:
     cdef Face _face
     cdef FontFuncs _ffuncs
 
-    def __cinit__(self, Face face):
+    def __init__(self, face_or_font):
```
anthrotype, line 398: allocation of C data structures should be done in the special `__cinit__` method, which is guaranteed to be called exactly once: 
https://cython.readthedocs.io/en/latest/src/userguide/special_methods.html#initialisation-methods-cinit-and-init 

i'll fix that on `main`

anthrotype: https://github.com/harfbuzz/uharfbuzz/commit/b7a809ddedd138622ccb2d9c573a64528a16eb67

Issue #102, opened 2021-08-19
Bind hb-subset


khaledhosny:
It would be nice to have bindings for the subset library, after its API is stabilized (i.e. after HarfBuzz 3.0.0).

madig:
If one were to hypothetically look into this, where would they start? Which functions and associated types, etc. need to be exposed?

behdad:
> If one were to hypothetically look into this, where would they start? Which functions and associated types, etc. need to be exposed?

https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset.h

behdad:
> If one were to hypothetically look into this, where would they start? Which functions and associated types, etc. need to be exposed?

The main entry point is `hb_subset_or_fail`:
https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-subset.h#L195

PR #142, opened 2022-12-31
Bind hb-subset API

madig:
Am I doing this right? Seems like I need to port over HB's set and map type. As functions or as standalone types? Hm...

Closes https://github.com/harfbuzz/uharfbuzz/issues/102


behdad:
One thing missing is in that file I see these lines:
```
cdef extern from "hb-ot.h": 
```

Next step would be to bind the set and map into higher-level objects I think.

behdad:
> Next step would be to bind the set and map into higher-level objects I think.

That can be done slowly; doesn't need to implement the full API initially IMO.

madig:
> One thing missing is in that file I see these lines:

What do you mean?

I bound the functions now (structs can remain opaque I guess). I'll doodle some test code now to see how this stuff fits together :eyes:

behdad:
I wonder if you should just compile harfbuzz-subset.cc instead.

madig:
Just adding `harfbuzz/src/hb-subset.cc` makes compilation fail because it can't find some CFF stuff, so I added everything. I now have a 41 MB .so file :eyes:

behdad:
> Just adding `harfbuzz/src/hb-subset.cc`

harfbuzz-subset.cc, not hb-subset.cc.

behdad:
> harfbuzz-subset.cc

That would be instead of everything.

madig:
That doesn't seem to work on Win and Mac? I'm probably doing it wrong.

Anyway, the following code seems to do something:

```cython
def test_subsetting(face: Face) -> Face:
    cdef hb_subset_input_t* input = hb_subset_input_create_or_fail()
    if input is NULL:
        raise MemoryError()

    cdef hb_face_t* preprocessed_face = hb_subset_preprocess(face._hb_face)
    
    cdef hb_set_t* gids = hb_subset_input_glyph_set(input)
    hb_set_add(gids, 0)
    hb_set_add(gids, 1)
    hb_set_add(gids, 9)
    hb_set_add(gids, 10)
    hb_subset_input_pin_axis_location(input, preprocessed_face, hb_tag_from_string(b"wght", -1), 900.0)
    hb_subset_input_pin_axis_location(input, preprocessed_face, hb_tag_from_string(b"wdth", -1), 125.0)
    hb_subset_input_pin_axis_location(input, preprocessed_face, hb_tag_from_string(b"ital", -1), 0)

    cdef hb_face_t* new_face = hb_subset_or_fail(preprocessed_face, input)
    new_face_inst = Face(None)
    new_face_inst._hb_face = new_face

    return new_face_inst
```

```py
from __future__ import annotations

import uharfbuzz as hb


def test_subsetting() -> None:
    path = "/home/me/Downloads/Mona.Sans.1.0/Mona Sans/Mona-Sans.ttf"
    blob = hb.Blob.from_file_path(path)
    face = hb.Face(blob)

    new_face = hb.test_subsetting(face)
    new_blob = hb.Blob.from_face(new_face)
    with open("/tmp/a.ttf", "wb") as f:
        f.write(new_blob.data)
```

So... any API wishes? Maybe a SubsetInput class that wraps `hb_subset_input_t` that you can add GIDs to and pin locations in? Is `hb_subset_plan_t` a superset of `hb_subset_input_t` and if so, should I wrap that instead? 

Aside: For instance cutting purposes, I'd like to have some type that says "keep everything put pin these locations". For debris-cleaning purposes, I'd like to also have some way of saying "Keep all glyphs that are reachable by cmap, composites and OTL, but drop the rest".

Aside 2: manual memory management, ugh.

behdad:
> That doesn't seem to work on Win and Mac? I'm probably doing it wrong.

Humm. Nothing jumps at me.

> So... any API wishes? Maybe a SubsetInput class that wraps `hb_subset_input_t` that you can add GIDs to and pin locations in? Is `hb_subset_plan_t` a superset of `hb_subset_input_t` and if so, should I wrap that instead?

No, a subset_plan is different object. cc @garretrieger 



> Aside: For instance cutting purposes, I'd like to have some type that says "keep everything put pin these locations".

For that get the hb_subset_input_glyph_set() and invert the set, to mean that you want to keep every glyph.

> For debris-cleaning purposes, I'd like to also have some way of saying "Keep all glyphs that are reachable by cmap, composites and OTL, but drop the rest".

For that get the hb_subset_input_unicode_set() and invert the set, to mean that you want to keep every Unicode.

behdad:
Also, preprocessing the face slows down subsetting. You should only do it if you are going to subset many times.

behdad:
> Aside: For instance cutting purposes, I'd like to have some type that says "keep everything put pin these locations". For debris-cleaning purposes, I'd like to also have some way of saying "Keep all glyphs that are reachable by cmap, composites and OTL, but drop the rest".

To truly keep everything is a bit more involved. We should make it easier. cc @garretrieger 
See https://github.com/harfbuzz/harfbuzz/blob/4adc748b13c3fdcb60162a269982590925750ce3/src/hb-subset-input.cc#L479-L515

behdad:
> > Aside: For instance cutting purposes, I'd like to have some type that says "keep everything put pin these locations". For debris-cleaning purposes, I'd like to also have some way of saying "Keep all glyphs that are reachable by cmap, composites and OTL, but drop the rest".
> 
> To truly keep everything is a bit more involved. We should make it easier. cc @garretrieger See https://github.com/harfbuzz/harfbuzz/blob/4adc748b13c3fdcb60162a269982590925750ce3/src/hb-subset-input.cc#L479-L515

I made that code a bit less copy-pasta:
https://github.com/harfbuzz/harfbuzz/blob/52110f13b02678c24daa3c1b588683a8fb13e125/src/hb-subset-input.cc#L485-L503

behdad:
> > Aside: For instance cutting purposes, I'd like to have some type that says "keep everything put pin these locations". For debris-cleaning purposes, I'd like to also have some way of saying "Keep all glyphs that are reachable by cmap, composites and OTL, but drop the rest".
> 
> To truly keep everything is a bit more involved. We should make it easier. cc @garretrieger See https://github.com/harfbuzz/harfbuzz/blob/4adc748b13c3fdcb60162a269982590925750ce3/src/hb-subset-input.cc#L479-L515

I added hb_subset_input_keep_everything() to main.

khaledhosny:
> That doesn't seem to work on Win and Mac?

The harfbuzz-subset.cc file misses hb-ft, hb-coretext etc, probably because they are not used by the subset library.

behdad:
> > That doesn't seem to work on Win and Mac?
> 
> The harfbuzz-subset.cc file misses hb-ft, hb-coretext etc, probably because they are not used by the subset library.

Nothing in his code uses those though. It's pure HarfBuzz code.

madig:
But why link against dwrite and ApplicationServices then?

madig:
Well duh, if access to system shapers is intended, I can't just source harfbuzz-subset.cc and will have to revert https://github.com/harfbuzz/uharfbuzz/pull/142/commits/6d2514b5f99ade58b1e4ec9e74753345330e3939 I think?

setup.py
```diff
@@ -43,7 +43,13 @@
...
         'src/uharfbuzz/_harfbuzz.pyx',
         'harfbuzz/src/harfbuzz.cc',
+        'harfbuzz/src/hb-subset-cff-common.cc',
+        'harfbuzz/src/hb-subset-cff1.cc',
+        'harfbuzz/src/hb-subset-cff2.cc',
+        'harfbuzz/src/hb-subset-input.cc',
+        'harfbuzz/src/hb-subset-plan.cc',
         'harfbuzz/src/hb-subset-repacker.cc',
+        'harfbuzz/src/hb-subset.cc',
         'harfbuzz/src/graph/gsubgpos-context.cc',
```
khaledhosny, line 53: You can do it the otherway, use `harfbuzz-subset.cc` + hb-coretext.cc, these platform integration files are unlikely to change.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -306,10 +306,26 @@ cdef class Blob:
         inst._hb_blob = blob
         return inst
 
+    @classmethod
+    def from_face(cls, face: Face) -> Blob:
```
khaledhosny, line 310: That is an odd place to have it, it should be a  Face method.

madig: Oh, hm, okay. The purpose is to get access to a blob to write to disk. What's the best way?

madig:
I'm having trouble deciding what to do with `hb_subset_preprocess`. I was thinking of making it a Face method, but am unsure about memory lifetimes... Does it allocate a new `hb_face_t` or does it modify it in place? I suppose I need to generate a new blob?

behdad:
> I'm having trouble deciding what to do with `hb_subset_preprocess`. I was thinking of making it a Face method

I think it should be a standalone function.

> but am unsure about memory lifetimes... Does it allocate a new `hb_face_t` or does it modify it in place? I suppose I need to generate a new blob?

It generates a new face.

madig:
The docs say it might keep old blobs around, so I can't just dealloc the old face and put in a new one, no?

behdad:
> The docs say it might keep old blobs around, so I can't just dealloc the old face and put in a new one, no?

You can. The blob is internally referenced.

pyproject.toml
```diff
@@ -16,3 +16,8 @@ test-command = "pytest {project}/tests"
 
 [tool.cibuildwheel.macos]
 archs = ["x86_64", "universal2", "arm64"]
+
+# Reduce the size of Linux wheels.
+[[tool.cibuildwheel.overrides]]
+select = "*-manylinux_*"
+environment = { CXXFLAGS="-g0" }
```
behdad, line 23: Push this as a separate PR?

madig: Yes, I was testing stuff because this PR is very WIP-y as is 😁

PR #143, opened 2023-01-01
Optimize draw_glyph_with_pen() to reuse draw funcs

behdad:
This is how draw_funcs are designed to be used.


behdad:
Khaled pointed out https://github.com/harfbuzz/uharfbuzz/pull/134

If we want to avoid the extra attribute lookup in the callbacks, we can put the bound methods in a struct and pass that as draw_data.

behdad:
> Khaled pointed out #134
> 
> If we want to avoid the extra attribute lookup in the callbacks, we can put the bound methods in a struct and pass that as draw_data.

Done.

behdad:
This now should be a net performance win.

justvanrossum (approved):

khaledhosny (approved):

PR #144, opened 2023-01-03
Add subsetter v2

madig:
Continuation of https://github.com/harfbuzz/uharfbuzz/pull/142 on an upstream branch.

Closes https://github.com/harfbuzz/uharfbuzz/issues/102.


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -392,7 +392,7 @@ cdef class Face:
     def upem(self, value: int):
         hb_face_set_upem(self._hb_face, value)
 
-    def reference_blob(self) -> Blob:
+    def get_blob(self) -> Blob:
```
khaledhosny, line 395: I’d even call it `blob` and make it a property.

behdad: Sounds good.

behdad:
I think this can be merged. In a separate PR I want to redo the memory-management to use HB's set_user_data facility, also see if blobs can be made cheaper by not making a bytes copy into Python as well.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -379,6 +392,18 @@ cdef class Face:
...
+    @property
+    def reference_blob(self) -> Blob:
+        cdef hb_blob_t* blob = hb_face_reference_blob(self._hb_face)
+        if blob is NULL:
+            raise MemoryError()
+        cdef Blob inst = Blob(None)
+        inst._hb_blob = blob
+        cdef unsigned int blob_length
+        cdef const_char* blob_data = hb_blob_get_data(blob, &blob_length)
+        inst._data = blob_data[:blob_length]
```
khaledhosny, line 404: Why is this done here and not in `Blob.data()`? may be also add `Blob.from_ptr()` and simplify the rest of this?

behdad: I agree.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1394,368 @@ def repack_with_tag(tag: str,
...
+
+    @property
+    def unicode_set(self) -> Set:
+        return Set.from_ptr(hb_set_reference (hb_subset_input_unicode_set(self._input)))
+
+    @property
+    def glyph_set(self) -> Set:
+        return Set.from_ptr(hb_set_reference (hb_subset_input_glyph_set(self._input)))
+
+    def set(self, set_type : SubsetInputSets) -> Set:
```
khaledhosny, line 1461: I think `set` is confusing here, I’d expect a method that allows me to set stuff. What if this became a dict-like property, e.g. `name_ids_set = input.sets[SubsetInputSets.NAME_ID]` or something like that, or even `name_ids_set = input[SubsetInputSets.NAME_ID]` and using a custom `__getitem__` w/o `__setitem__`?

behdad: I like the __getitem__ idea.  __setitem__ can also be implemented by overriding that entire set.  Let me implement that.

behdad: I'll keep the name `sets` internally.

behdad: Done.

behdad: @madig had originally added properties for each of them individually. That's also an option; I wasn't a fan of it...

khaledhosny: Yes, I saw it. I think the current solution is better as it requires less maintenance going forward.

khaledhosny: But I think If we really want to do this, we can use `__getattr__`/`__setattr__`, so that something like `input.name_id` translates to `input[SubsetInputSets.NAME_ID]`.

khaledhosny: or `input.name_id_set`

behdad: Yeah I thought about that too. That sounds like too much magic. I'm happy with current solution I think; either this or hand-written properties. getattr/setattr goes too far for me for most cases.

behdad: Maybe we go back to what madig had?

behdad: I mean, we keep the sets() or set() API, but also add properties? In HB API we specifically didn't add those. Again, I think I prefer the getitem/setitem and call it a day.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+    DEFAULT = HB_SUBSET_FLAGS_DEFAULT
+    NO_HINTING = HB_SUBSET_FLAGS_NO_HINTING
+    RETAIN_GIDS = HB_SUBSET_FLAGS_RETAIN_GIDS
+    DESUBROUTINIZE = HB_SUBSET_FLAGS_DESUBROUTINIZE
+    NAME_LEGACY = HB_SUBSET_FLAGS_NAME_LEGACY
+    SET_OVERLAPS_FLAG = HB_SUBSET_FLAGS_SET_OVERLAPS_FLAG
+    PASSTHROUGH_UNRECOGNIZED = HB_SUBSET_FLAGS_PASSTHROUGH_UNRECOGNIZED
+    NOTDEF_OUTLINE = HB_SUBSET_FLAGS_NOTDEF_OUTLINE
+    GLYPH_NAMES = HB_SUBSET_FLAGS_GLYPH_NAMES
+    NO_PRUNE_UNICODE_RANGES = HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES
```
khaledhosny, line 1491: Can we move this up next to `SubsetInputSets`, it looks odd with one before the `SubsetInput` class and one after it.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+            raise MemoryError()
+
+        self.set(init)
+
+    def __dealloc__(self):
+        hb_set_destroy(self._hb_set)
+
+    @staticmethod
+    cdef Set from_ptr(hb_set_t* hb_set):
+        """Create Set from taking ownership of a pointer."""
```
khaledhosny, line 1511:
```suggestion
        """Create Set from a pointer taking ownership of it."""
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+
+        for k,v in init.items():
+            self.set(k, v)
+
+    def __dealloc__(self):
+        hb_map_destroy(self._hb_map)
+
+    @staticmethod
+    cdef Map from_ptr(hb_map_t* hb_map):
+        """Create Map from taking ownership of a pointer."""
```
khaledhosny, line 1698:
```suggestion
        """Create Map from a pointer taking ownership of it."""
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -288,14 +288,26 @@ cdef class Blob:
...
             self._data = data
             self._hb_blob = hb_blob_create(
                 data, len(data), HB_MEMORY_MODE_READONLY, NULL, NULL)
         else:
+            self._data = bytes()
             self._hb_blob = hb_blob_get_empty()
 
+    @staticmethod
+    cdef Blob from_ptr(hb_blob_t* hb_blob):
+        """Create Blob from taking ownership of a pointer."""
```
khaledhosny, line 302:
```suggestion
        """Create Blob from a pointer taking ownership of it."""
```

src/uharfbuzz/_harfbuzz.pyx
khaledhosny: Agree, it is good the way it is now.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -345,11 +361,20 @@ cdef class Face:
...
             self._hb_face = hb_face_get_empty()
+            self._blob = None
 
     def __dealloc__(self):
         hb_face_destroy(self._hb_face)
         self._blob = None
 
+    @staticmethod
+    cdef Face from_ptr(hb_face_t* hb_face):
+        """Create Face from taking ownership of a pointer."""
```
khaledhosny, line 372:
```suggestion
        """Create Face from a pointer taking ownership of it."""
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+        return hb_map_hash(self._hb_map)
+
+    def __hash__(self) -> int:
+        return self.hash()
+
+    def set(self, k: int, v: int):
+        hb_map_set(self._hb_map, k, v)
+        if not hb_map_allocation_successful(self._hb_map):
+            raise MemoryError()
+
```
khaledhosny, line 1742: Do we want also `update()` like dict? May be also `keys()`, `values()` etc.?

khaledhosny:
Do we want also a `SubsetPlan` class?

src/uharfbuzz/_harfbuzz.pyx
behdad: `update()` sounds good from a Python `dict` but we don't have API to update from a HB Dict. We don't have HB API for `keys()` and `values()`.

behdad:
> Do we want also a `SubsetPlan` class?

Yes; I'll work on that.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+    @staticmethod
+    cdef Set from_ptr(hb_set_t* hb_set):
+        """Create Set from a pointer taking ownership of a it."""
+
+        cdef Set wrapper = Set.__new__(Set)
+        wrapper._hb_set = hb_set
+        return wrapper
+
+    def copy(self) -> Set:
```
madig, line 1517: This should probably be a dunder method: https://devdocs.io/python~3.8/library/copy

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+        c = Set()
+        c._hb_set = hb_set_copy(self._hb_set)
+        return c
+
+    def clear(self):
+        hb_set_clear(self._hb_set)
+
+    def is_empty(self) -> bool:
```
madig, line 1525: Idiomatically, I think this should be just `__bool__`?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+        return hb_set_is_empty(self._hb_set)
+
+    def __bool__(self) -> bool:
+        return not self.is_empty()
+
+    def invert(self):
+        hb_set_invert(self._hb_set)
+
+    def has(self, c: int) -> bool:
```
madig, line 1534: Idiomatically, I think this should be just `__contains__`?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+        hb_set_add(self._hb_set, c)
+        if not hb_set_allocation_successful(self._hb_set):
+            raise MemoryError()
+
+    def add_range(self, first: int, last: int):
+        hb_set_add_range(self._hb_set, first, last)
+        if not hb_set_allocation_successful(self._hb_set):
+            raise MemoryError()
+
+    def remove(self, c: int):
```
madig, line 1552: Idiomatically, I think this should be just `__delitem__`?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+
+    def del_range(self, first: int, last: int):
+        hb_set_del_range(self._hb_set, first, last)
+        if not hb_set_allocation_successful(self._hb_set):
+            raise MemoryError()
+
+    def __delitem__(self, c: int):
+        self.remove(c)
+
+    def is_equal(self, other: Set) -> bool:
```
madig, line 1572: Idiomatically, I think this should be just `__eq__`?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,375 @@ def repack_with_tag(tag: str,
...
+        return hb_set_is_equal(self._hb_set, other._hb_set)
+
+    def __eq__(self, other):
+        if type(other) != Set:
+            return NotImplemented
+        return self.is_equal(other)
+
+    def hash(self) -> int:
```
madig, line 1580: Idiomatically, I think this should be just `__hash__`? `hash` is a builtin.

madig:
I'm strongly in favor over only implementing dunder methods for copying, setting, getting, etc., to keep it idiomatic.

src/uharfbuzz/_harfbuzz.pyx
behdad: Added. Keeping copy() because both set() and dict() have that.

src/uharfbuzz/_harfbuzz.pyx
behdad: set() also has remove and discard.

src/uharfbuzz/_harfbuzz.pyx
behdad: > I'm strongly in favor over only implementing dunder methods for copying, setting, getting, etc., to keep it idiomatic in Python.

Done.

> Also, I'd suggest keeping the `property` getters and setters for SubsetInput, because it allows us to ignore `SubsetInputSets` in the public API :)

I'm not convinced that's a good thing though. But sure, go ahead and add those on top.

src/uharfbuzz/_harfbuzz.pyx
justvanrossum: Also: set doesn't have `__delitem__`

behdad:
> Do we want also a `SubsetPlan` class?

Done.

behdad:
This is ready from my points of view.

> Also, I'd suggest keeping the `property` getters and setters for SubsetInput, because it allows us to ignore `SubsetInputSets` in the public API :)

madig still wants this.

src/uharfbuzz/_harfbuzz.pyx
behdad: Done.

behdad:
> > Also, I'd suggest keeping the `property` getters and setters for SubsetInput, because it allows us to ignore `SubsetInputSets` in the public API :)
> 
> madig still wants this.

Done.

behdad:
Added map iterator, keys(), values(), and items() using a new API in harfbuzz main.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,437 @@ def repack_with_tag(tag: str,
...
+    NAME_LEGACY = HB_SUBSET_FLAGS_NAME_LEGACY
+    SET_OVERLAPS_FLAG = HB_SUBSET_FLAGS_SET_OVERLAPS_FLAG
+    PASSTHROUGH_UNRECOGNIZED = HB_SUBSET_FLAGS_PASSTHROUGH_UNRECOGNIZED
+    NOTDEF_OUTLINE = HB_SUBSET_FLAGS_NOTDEF_OUTLINE
+    GLYPH_NAMES = HB_SUBSET_FLAGS_GLYPH_NAMES
+    NO_PRUNE_UNICODE_RANGES = HB_SUBSET_FLAGS_NO_PRUNE_UNICODE_RANGES
+
+
+cdef class SubsetInput:
+    cdef hb_subset_input_t* _input
```
khaledhosny, line 1439:
```suggestion
    cdef hb_subset_input_t* _hb_input
```

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1369,3 +1401,437 @@ def repack_with_tag(tag: str,
...
+        cdef unsigned subset_flags = hb_subset_input_get_flags(self._input)
+        return SubsetFlags(subset_flags)
+
+    @flags.setter
+    def flags(self, flags: SubsetFlags) -> None:
+        hb_subset_input_set_flags(self._input, int(flags))
+
+
+cdef class SubsetPlan:
+    cdef hb_subset_plan_t* _plan
```
khaledhosny, line 1514:
```suggestion
    cdef hb_subset_plan_t* _hb_plan
```

khaledhosny (approved):

behdad:
Thanks Khaled.

PR #145, opened 2023-01-03
Strip debug info on Linux

madig:
Reduces wheel size.


behdad:
LGTM.

madig:
Wait, this doesn't seem to do anything actually :( https://github.com/harfbuzz/uharfbuzz/actions/runs/3832875123/jobs/6523690530#step:5:1919

PR #146, opened 2023-01-04
Try and speed up Linux builds and reduce wheel size

madig:
Continuation of #145.

Also try ccaching HarfBuzz compilation.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,8 +16,19 @@ test-command = "pytest {project}/tests"
 
 [tool.cibuildwheel.macos]
 archs = ["x86_64", "universal2", "arm64"]
+before-all = "brew install ccache"
+before-test = "ccache --show-stats"
 
-# Reduce the size of Linux wheels.
-[[tool.cibuildwheel.overrides]]
-select = "*-manylinux_*"
-environment = { CXXFLAGS="-g0" }
+[tool.cibuildwheel.macos.environment]
+CC = "/usr/local/opt/ccache/libexec/cc"
+
+[tool.cibuildwheel.linux]
+before-all = [
+    "yum install -y ccache",
+    "ln -s $(which ccache) /usr/lib64/ccache/gcc",
+]
+before-test = "ccache --show-stats"
+
+[tool.cibuildwheel.linux.environment]
+CC = "/usr/lib64/ccache/gcc"
+CFLAGS = "-g0"


madig:
cibuildhweel seems to call gcc rather than g++ to compile stuff?

pyproject.toml
```diff
@@ -17,7 +17,18 @@ test-command = "pytest {project}/tests"
...
-select = "*-manylinux_*"
-environment = { CXXFLAGS="-g0" }
+# Try caching compilation of harfbuzz.
+[tool.cibuildwheel.linux]
+before-build = [
+    "yum install -y ccache",
+    "echo '#!/bin/sh\nccache gcc $@' > /usr/local/bin/gcc-ccache",
+    "chmod +x /usr/local/bin/gcc-ccache",
+    "echo '#!/bin/sh\nccache g++ $@' > /usr/local/bin/g++-ccache",
+    "chmod +x /usr/local/bin/g++-ccache",
```
khaledhosny, line 27: In Debian, there is a dir you add to the PATH after installing ccache so that cc, gcc etc are all symlinks to ccache. Fedora/RH probably have something similar.

madig: Ah, I see. The directory seems to be empty though after installation on the manylinux 2014 image?!

khaledhosny: I don’t know, but if this works then 🤷🏽‍♂️

madig: I'm trying to do better but it seems to be slower now, got 60s per wheel build at some point and now it's 90s. Not sure if this is the config or GHA being slower for some reason...

behdad:
@madig Your cat is on the keyboard.

madig:
I am documenting my frustration with debugging things on a CI.

madig:
Finally. No idea what's goingon with PATH, but specifying ccache directly does the trick. Or something.

Next step: use ccache on macOS. And Windows maybe?

behdad:
> And Windows maybe?

That sounds next-level.

madig:
https://github.com/ccache/ccache/wiki/MS-Visual-Studio

khaledhosny:
Can we also cache the ccache cache and use it across the ~cache~ci builds.

madig:
Behdad mentioned that for regular CI, we can just test min and max supported version instead of everything.

khaledhosny:
This is already a huge improvement, I’d merge it as it is and do more in another PR.

madig:
The improvements on macOS and Windows seem to vary greatly. Not sure sure why. Is the cache even used on Windows?!

khaledhosny:
> The improvements on macOS and Windows seem to vary greatly. Not sure sure why. Is the cache even used on Windows?!

I thought I was seeing macOS builds around 20 minutes, but now I see it is 45 minutes which is still too slow indeed.

madig:
Yes, they seem to depend on the time of day or something.

madig:
Marking this ready because it's better than nothing. Windows and breaking up the CI runs can be done later...

anthrotype:
to reduce macOS build time, I think we don't need to build both arm64 and universal2 wheels (in addition to x86_64 wheels), we can get away with only building universal2 (covering both x86_64 and arm64) and x86_64 for legacy stuff that doesn't understand universal2

madig:
Or x86_64 and arm64 wheels separately, no universal2?

PR #148, opened 2023-01-05
Drop `Set.__repr__`

madig:
An infinite set loops forever.


behdad:
Sounds good to me. I might find something better later.

behdad:
Better fix: https://github.com/harfbuzz/uharfbuzz/pull/149

PR #149, opened 2023-01-06
Set: Fix repr for inverted sets

behdad:


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1843,8 +1846,11 @@ cdef class Set:
         return SetIter(self)
 
     def __repr__(self):
-        s = ', '.join(repr(v) for v in self)
-        return ("Set({%s})" % s)
+        if self.is_inverted():
+            return "Set({...})"
+        else:
```
madig, line 1851: No `else` needed here because the above returns :)

madig (approved):

PR #156, opened 2023-01-21
Play with meson

madig:
This almost does something.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,25 @@
+project('foo', 'cython', 'cpp')
+
+py = import('python').find_installation()
+dep_py = py.dependency()
+
+harfbuzz_includes = include_directories('harfbuzz/src')
+harfbuzz = static_library(
+    'harfbuzz',
+    'harfbuzz/src/harfbuzz-subset.cc',
+    'harfbuzz/src/hb-coretext.cc',
+    'harfbuzz/src/hb-directwrite.cc',
+    'harfbuzz/src/hb-uniscribe.cc',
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,11 +1,14 @@
 [build-system]
+build-backend = "mesonpy"
 requires = [
-    "setuptools >= 36.4",
-    "wheel",
+    "meson-python",
```
khaledhosny, line 4: This should be `meson`, and you probably want `ninja` too.

meson.build
```diff
@@ -0,0 +1,25 @@
...
+harfbuzz_includes = include_directories('harfbuzz/src')
+harfbuzz = static_library(
+    'harfbuzz',
+    'harfbuzz/src/harfbuzz-subset.cc',
+    'harfbuzz/src/hb-coretext.cc',
+    'harfbuzz/src/hb-directwrite.cc',
+    'harfbuzz/src/hb-uniscribe.cc',
+    cpp_args: '-DHB_NO_MT=1 -DHB_EXPERIMENTAL_API=1',
+    include_directories: harfbuzz_includes,
+)
```
khaledhosny, line 15: If you are going to use meson, why not use harfbuzz as a subproject (or simply as a [dependency](https://mesonbuild.com/Reference-manual_functions.html#dependency), so that #45 could be fixed as well). You can

pyproject.toml
khaledhosny: Oh, meson-python is really a thing!

meson.build
madig: Yes, that's the idea, but first I want to get it to actually compile. It still complains about missing hb.h.

PR #162, opened 2023-07-09
Update HarfBuzz to 8.0.0

behdad:


khaledhosny:
We should try to enable the wasm shaper. cc @simoncozens.

PR #165, opened 2023-07-18
add explicit noexcept to callbacks used by hb C functions

### LINKED ISSUE #164: can't install uharfbuzz due to cython new version 3
pip install can't work due to cython 3.0 release
I have try to correct this in my PR
https://github.com/harfbuzz/uharfbuzz/pull/163

anthrotype:
Fixes #164

with this change I can build uharbfbuzz also with the latest Cython 3.0. It continues to build with 0.29 as well.

The 'noexcept' is now required for non-extern cdef functions that are otherwise assumed to propagate exceptions by default in Cython 3 (previously they'd only do if `except` was used). Some cdef functions in uharfbuzz are passed as callbacks to C hb functions where python exceptions can't be propagated as far as I understand (or can they?), so the fix was to explicitly mark all these with `noexcept` and then Cython is happy again.
I think this is safe and doesn't change the previous behavior since we had not marked these as `except` before.

See https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html?highlight=noexcept#exception-values-and-noexcept

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -30,7 +30,7 @@ cdef extern from "Python.h":
     int PyUnicode_4BYTE_KIND
 
 
-cdef int msgcallback(hb_buffer_t *buffer, hb_font_t *font, const char* message, void* userdata):
+cdef int msgcallback(hb_buffer_t *buffer, hb_font_t *font, const char* message, void* userdata) noexcept:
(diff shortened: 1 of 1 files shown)


behdad:
LGTM.

fredericoschardong:
Can we have a minor release of `uharfbuzz` with this fix for Cython 3.0? Other projects such as [pyHanko](https://github.com/MatthiasValvekens/pyHanko/issues/287) are depending on it.

PR #167, opened 2023-07-19
Build wheels for musl-based linux

### LINKED ISSUE #166: Wheels for alpine/musl
Some of us out there are using Python on alpine-based systems, which use musl instead of glibc. Thus, we have to build our own wheels everytime.

Can you please add musl-based wheels? It would imensely speed up the build process.

fredericoschardong:
It is still faster to build wheels for both `manylinux` and `musllinux` than windows and macos:

image

Close #166

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ name: Build + Deploy
 on:
   push:
     branches: [main]
-    tags: ["v*.*.*"]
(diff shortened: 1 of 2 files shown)


anthrotype:
Thank you!

anthrotype:
just pushed v0.37.1.post1 git tag, the wheels should be on PyPI in half hour or so

Issue #168, opened 2023-07-20
Crash when repacking fonts with latest binary wheels (v0.37.1, v0.37.1.post1)


Hoolean:
Hello!

A segfault occurs when building fonts on Windows (amd64) with the latest uharfbuzz binary wheels ([v0.37.1](https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.37.1), [v0.37.1.post1](https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.37.1.post1)) on pypi. More specifically, it occurs when saving TTFs with HarfBuzz binary repacking enabled.

## Reproducing

1. Clone googlefonts/roboto-classic@58302217
2. Create a venv and install fontmake and uharfbuzz==0.37.1 (Windows amd64 binary wheel)
3. Run `fontmake -m .\sources\Roboto.designspace -o variable`
4. fontmake segfaults, the output log is truncated, and no TTF is created

## Additional Details

In addition to testing with the binary wheel deployed on pypi, we also tested locally, which gave strange results:

| uharfbuzz version | Windows 10 | Windows 11 |
|-|-|-|
| v0.37.0 amd64 pypi wheel | ✔️ | ✔️ |
| v0.37.1 amd64 pypi wheel | 💥 | 💥 |
| v0.37.1.post1 amd64 pypi wheel | 💥 | 💥 |
| v0.37.1.post1 built locally with VS Community 2022 v17.6.4 | ✔️ | ❔ |
| v0.37.1.post1 built locally with VS Community (or Build Tools?) 2022 v17.5.5 | ❔ | 💥 |
| v0.37.1.post1 with HarfBuzz 7 retrofitted, built locally with VS Community (or Build Tools?) 2022 v17.5.5 | ❔ | ✔️ |
| v0.37.1.post1 built locally with VS Community 2022 v17.6.5 | ❔ | ✔️ |

behdad:
@garretrieger

behdad:
Seems to work for me on Linux.

Hoolean:
Whoops, yes, I forgot to add that we tested one of the Linux pypi wheels for v0.37.1.post1 in WSL2 and it worked fine :+1:

Hoolean:
From a hunt with @anthrotype and @madig, we suspect that you can reproduce reliably with a uharfbuzz based on HarfBuzz 8 and an unlucky Windows build environment (including the one used in this project's CI runner)

behdad:
[...]

PR #170, opened 2023-07-27
More verbosity and AddressSanitizer

Hoolean:
For isolating #168, supersedes #169

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [windows-latest]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Hoolean:
For future debuggers: this action [built a binary wheel](https://github.com/harfbuzz/uharfbuzz/actions/runs/5679949990/job/15393095807) of uharfbuzz with ASan included, and this can be run locally with some extra wangling to make sure ASan is available when the extension is loaded:
* https://stackoverflow.com/a/75380589
* https://stackoverflow.com/a/67437837

HinTak:
That's not surprising - crashes often disappear when recompiled, particularly if you try to compile in an address sanitizer/ memory debugger. Speaking of which, I have seen crashes disappearing recently with ft2-demos when run under Valgrind - Valgrind cleans up the heap so it doesn't crash, but it has indicated where a crash "could have been". Can you run valgrind under windows in "capture all forks" mode (this is for scripts etc where a script - in my case, libtool script, runs binaries)?

Hoolean:
We are safe from #168! <sup>for now</sup>

Closing this PR.

Thanks for the suggestion too HinTak :)

PR #171, opened 2023-07-28
Test

behdad:


behdad:
@anthrotype Is there a way to tell cibuildwheel to only build for one target, so I can get faster builds? Currently it takes a looong time to finish all the various wheel builds before the artifact I want gets uploaded...

behdad:
`CIBW_BUILD` perhaps.

PR #177, opened 2023-10-18
WIP: make tracing more usable

khaledhosny:


behdad:
LGTM. THanks.

khaledhosny:
Thinking about it more, this is probably not needed. One has access to the Python Buffer object when setting the message function and can use a pointer to it when defining the function, which avoids the hb_buffer_reference()/hb_buffer_destroy() call for each message function call.

PR #178, opened 2023-11-17
Add `font.glyph_to_string` to README.md

roberto-arista:


anthrotype (approved):

anthrotype:
there seems to be two methods that do the same thing (convert a gid to a glyph name), i.e. hb_font_get_glyph_name and hb_font_glypy_to_string, I wonder why and what's the diff /cc @khaledhosny @behdad

khaledhosny:
> there seems to be two methods that do the same thing (convert a gid to a glyph name), i.e. hb_font_get_glyph_name and hb_font_glyph_to_string, I wonder why and what's the diff /cc @khaledhosny @behdad

`hb_font_glyph_to_string()` always returns a string and will synthesize a name (`gidNNN`) if there is no glyph name, while `hb_font_get_glyph_name()` only returns glyph names if present.

anthrotype:
ok thanks. So I suppose get_glyph_name() here is appropriate in the README and we can merge this?

khaledhosny:
For uses like this, `font.glyph_to_string()` is preferable, as `font.get_glyph_name()` can return None.

anthrotype:
right ok, @roberto-arista can you use glyph_to_string() instead? thanks!

roberto-arista:
done!

PR #184, opened 2024-01-04
Bind hb_ot_layout_lookup_get_glyph_alternates()

Tom-99:
Added the binding for hb_ot_layout_lookup_get_glyph_alternates() to allow querying the stylistic alternates of a glyph from Python.


khaledhosny (approved):
Please add also a test.

PR #185, opened 2024-01-20
Bind hb-paint API

khaledhosny:


behdad:
Nice. Thanks.

khaledhosny:
I’m starting to think the PyCapsule stuff is not of much use here, since the C side would need to link to HarfBuzz to be able to access the various C objects (`hb_font_t`, `hb_blob_t`, `hb_color_line_t`, and so on). So I think I’m going to rip it away.

The test failure is something I don’t understand. It seems that calling `PaintFuncs.set_color_glyph_func()` Somehow results in the `color_glyph` function being set to the `push_clip_glyph` function. There is probably some bug in my code, but I lost countless hours trying to find it in vain, that or I’m hitting some mysterious cython/python bug.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -839,6 +852,21 @@ cdef class Font:
             draw_state_p = <void *>PyCapsule_GetPointer(draw_state, NULL)
         hb_font_draw_glyph(self._hb_font, gid, draw_funcs._hb_drawfuncs, draw_state_p);
 
+    def paint_glyph(self, gid: int,
+                    paint_funcs: PaintFuncs,
+                    paint_state: object = None,
+                    palette_index: int = 0,
+                    foreground: int = 0x000000FF) -> None:
```
khaledhosny, line 859: Should we have a nicer representation of colors than raw integers?

behdad: Probably a named tuple?

behdad: I don't have a preference honestly.

behdad:
> The test failure is something I don’t understand. It seems that calling `PaintFuncs.set_color_glyph_func()` Somehow results in the `color_glyph` function being set to the `push_clip_glyph` function. There is probably some bug in my code, but I lost countless hours trying to find it in vain, that or I’m hitting some mysterious cython/python bug.

Did you figure this out?

khaledhosny:
> > The test failure is something I don’t understand. It seems that calling `PaintFuncs.set_color_glyph_func()` Somehow results in the `color_glyph` function being set to the `push_clip_glyph` function. There is probably some bug in my code, but I lost countless hours trying to find it in vain, that or I’m hitting some mysterious cython/python bug.
> 
> Did you figure this out?

No, but I did something different. I was passing the functions as user_data, but now I pass the PaintFuncs instance and access the functions from there.

PR #189, opened 2024-02-03
Don’t pass font to pain functions

khaledhosny:
Avoids the hb_(font|face)_(reference|destroy) at each function call. If the user needs the font, they can pass the Python Font in the paint_state.

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -1728,8 +1728,7 @@ cdef hb_bool_t _paint_color_glyph_func(
         hb_font_t *font,
         void *user_data) noexcept:
     py_funcs = <PaintFuncs>user_data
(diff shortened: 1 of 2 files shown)


khaledhosny:
Painting U+270D from tests/data/noto_handwriting-cff2_colr_1.otf using paint functions that does nothing:

Without this change:
```
real    0m1.441s
user    0m1.423s
sys     0m0.013s
```

With this change
```
real    0m1.269s
user    0m1.252s
sys     0m0.012s
```

behdad:
I was very worried before you said 100000 times :))

khaledhosny:
> I was very worried before you said 100000 times :))

:) I thought someone would be

behdad:
LGTM.

HinTak:
I was more worried about the mention of "pain" functions in the title. So harfbuzz is in the business of adding pain and suffering to the world now? 😉

PR #190, opened 2024-02-11
Add an example of using PaintFuncs with cairo

khaledhosny:


behdad:
Thanks. Maybe we ahould also bind hb-cairo?

khaledhosny:
> Maybe we ahould also bind hb-cairo?

I wrote this mainly to test the PaintFuncs bindings, then wanted to put it somewhere not not loose it. Binding hb-cairo would require also building cairo which I don’t think is going to be as easy as building HarfBuzz.

behdad:
> > Maybe we ahould also bind hb-cairo?
> 
> I wrote this mainly to test the PaintFuncs bindings, then wanted to put it somewhere not not loose it.

Sure.

> Binding hb-cairo would require also building cairo which I don’t think is going to be as easy as building HarfBuzz.

I was thinking providing support if pycairo is already available. I think that's possible without a build-time option.

PR #195, opened 2024-03-17
Create python-publish.yml

wexkalebur:


anthrotype:
Thanks but we already have something like this in our CI workflows.

PR #201, opened 2024-08-21
Use System HarfBuzz

### LINKED ISSUE #45: Option to link against installed harfbuzz

BwL1289:
Resolves #45.

Adds the ability to use system HarfBuzz and associated documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,20 @@ for info, pos in zip(infos, positions):
     print(f"{glyph_name} gid{gid}={cluster}@{x_advance},{y_offset}+{x_advance}")
 ```
 
+### Installation
+
+When building the uharfbuzz package, it automatically incorporates minimal HarfBuzz sources so you don't have to install the native HarfBuzz library.
+
+However, if you want to use uharfbuzz with your system-provided HarfBuzz (e.g., if you built it from sources with custom configuration), you can set `USE_SYSTEM_LIBS=1` environment variable (see example below).
+
+```shell
+USE_SYSTEM_LIBS=1 pip install uharfbuzz --no-binary :uharfbuzz:
+```
+
+harfbuzz installation is found using `pkg-config`, so you must have harfbuzz's `.pc` files in your system.
+If you've built it from sources, meson installs them automatically. Otherwise, you may want to install harfbuzz development package, like `harfbuzz-devel` on Fedora-derived distros.
+
+Note: you **must** build HarfBuzz with experimental API support enabled.
 
 ### How to make a release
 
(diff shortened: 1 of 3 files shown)


BwL1289:
@khaledhosny please let me know if this is acceptable.

CC @gorloffslava.

anthrotype:
thanks for this PR!
it's not clear to me why should pypy 3.8 fail on the CI but 3.7 still work.. From the error message it might even be some bug in setuptools and/or pypy itself.
It doesn't seem like you made any changes to our own CI workflow, it's still only building the uharfbuzz module with vendored harfbuzz, which was working before this PR, so it's not clear to me why it would start failing now.
I just pushed an empty commit to this https://github.com/harfbuzz/uharfbuzz/pull/202 (based off main) to see if it triggers there as well, without your changes.
Maybe we could also try to build with a more up to date PyPy version, if that works i'm happy to drop the ones that are currently failing.

BwL1289:
@anthrotype awesome - thanks so much for the help on this. really appreciate it.

anthrotype:
it looks like it failed with the same error so your PR is not the culprit
https://github.com/harfbuzz/uharfbuzz/actions/runs/10512034395/job/29124400429?pr=202#step:5:3264

I'd say let's not block this PR on the PyPy issue.

I was wondering, how easy or difficult would be to set up the CI to also test with USE_SYSTEM_LIBS=1? It'd be ok to only test this on linux

BwL1289:
> it looks like it failed with the same error so your PR is not the culprit https://github.com/harfbuzz/uharfbuzz/actions/runs/10512034395/job/29124400429?pr=202#step:5:3264
> 
> I'd say let's not block this PR on the PyPy issue.

Thank you!

> I was wondering, how easy or difficult would be to set up the CI to also test with USE_SYSTEM_LIBS=1? It'd be ok to only test this on linux

 I can put this in our backlog. We're juggling a lot with a small team (other contributions to tons of open source libs in addition to the platform we're building) so anything we can offload for now is greatly appreciated.

anthrotype:
thanks, I just tested this locally and it can confirm it works. I'll open an issue for exercising the USE_SYSTEM_LIBS=1 on our CI as well.

BwL1289:
Thanks @anthrotype

anthrotype:
https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.39.4

PR #202, opened 2024-08-22
pin setuptools to work around build error with PyPy 3

anthrotype:


anthrotype:
I was able to reproduce the `TypeError: 'NoneType' object is not subscriptable (key slice(None, None, None))` with PyPy 3.9 installed locally on my linux machine. It seems to be a bug or regression in setuptools' integration with pypy because I can trigger it with very simple extension modules.

anthrotype:
it looks like this only happens with setuptools >= 72.2 (current is 73.0.1). If we pin to setuptools<72.2 in pyproject.toml it works. I'll do that for now, and file a bug upstream tomorrow

anthrotype:
so yeah it's confirmed this is a regression in setuptools (actually in the vendored distutils module): https://github.com/pypa/distutils/issues/283

let's pin setuptools for the time being

PR #204, opened 2024-08-27
Add support for ot-metrics

skef:
Some test infrastructure I'm fooling with would benefit from this.

Hoping you might have some advice on whether or how I should test the x/y_variation functions, or if I should just delete those ...

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -1656,7 +1656,7 @@ class OTMetricsTag(IntEnum):
     UNDERLINE_OFFSET = HB_OT_METRICS_TAG_UNDERLINE_OFFSET
 
 def ot_metrics_get_position(font: Font,
-                            tag: OTMathTag) -> int:
+                            tag: OTMetricsTag) -> int:
     cdef hb_position_t hb_position
     cdef hb_bool_t success
     success = hb_ot_metrics_get_position(font._hb_font, tag, &hb_position)
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -1659,37 +1659,29 @@ def ot_metrics_get_position(font: Font,
                             tag: OTMetricsTag) -> int:
     cdef hb_position_t hb_position
     cdef hb_bool_t success
-    success = hb_ot_metrics_get_position(font._hb_font, tag, &hb_position)
-    if success:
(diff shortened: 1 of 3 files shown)


anthrotype:
thanks!

> some advice on whether or how I should test the x/y_variation functions, or if I should just delete those ...

what's different/special about those as far as testing is concerned?

skef:
> what's different/special about those as far as testing is concerned?

Looking at the source, they seem to scale the returned values based on `em_scalef_x` or `em_scalef_y`, which use scaling factors in the font. But I didn't dig too deeply on how to set those factors. @khaledhosny might have a better idea.

skef:
After thinking about this I added a mundane test for each of the methods. Given that uharfbuzz is a very thin layer over HarfBuzz, that's probably fine. Even if these methods aren't useful now, perhaps future additions will make them so.

khaledhosny:
I’m not familiar with this API, to be honest.

anthrotype:
fwiw i'd be fine with merghing this as is

skef:
If it passes review, including verifying that I'm calling each method I added in the tests in a somewhat substantive way, I think it will be fine.

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -1624,6 +1624,73 @@ def ot_math_get_glyph_assembly(font: Font,
...
+    SUPERSCRIPT_EM_Y_SIZE = HB_OT_METRICS_TAG_SUPERSCRIPT_EM_Y_SIZE
+    SUPERSCRIPT_EM_X_OFFSET = HB_OT_METRICS_TAG_SUPERSCRIPT_EM_X_OFFSET
+    SUPERSCRIPT_EM_Y_OFFSET = HB_OT_METRICS_TAG_SUPERSCRIPT_EM_Y_OFFSET
+    STRIKEOUT_SIZE = HB_OT_METRICS_TAG_STRIKEOUT_SIZE
+    STRIKEOUT_OFFSET = HB_OT_METRICS_TAG_STRIKEOUT_OFFSET
+    UNDERLINE_SIZE = HB_OT_METRICS_TAG_UNDERLINE_SIZE
+    UNDERLINE_OFFSET = HB_OT_METRICS_TAG_UNDERLINE_OFFSET
+
+def ot_metrics_get_position(font: Font,
+                            tag: OTMathTag) -> int:
```
khaledhosny, line 1659:
```suggestion
                            tag: OTMetricsTag) -> int:
```

skef:
Sorry, don't know where my head was on this one ...

khaledhosny (approved):

PR #209, opened 2024-09-15
Make hb-ot-* bindings more Pythonic

khaledhosny:
They were added as top level functions, but that feels odd on Python when we already have `Face` and `Font` classes. This deprecate the top level functions and adds corresponding methods to `Face` or `Font` classes as appropriate.


behdad:
LGTM. Thanks.

Issue #206, opened 2024-09-15
Documentation


khaledhosny:
We don’t seem to have documentation anywhere, but we should. Even though we try to remain close to HarfBuzz C-API as much as possible, it is still non 1:1 and I often find myself looking into the source code to find the right API to use.

I don’t know what options we have for documenting Cython code, so opening this in case someone have an idea.

madig:
I think Sphinx should be able to deal with Cython stuff as well. https://stackoverflow.com/questions/10057476/how-do-i-use-sphinx-with-cython says you may need to build the extension in place so that Sphinx can load and inspect it.

PR #212, opened 2024-09-22
Add API reference generated by Sphinx

Hoolean:
Hello - I've had a stab at #206!

This is largely a modification of `sphinx-quickstart` to automatically generate an API reference for the module, taking a few hints from fonttools/ufoLib2 to inform structure, build workflow, and styling.

It can be tested as follows:

```bash
tox -e docs
```

...and seems to be fairly happy with Cython, aside from a known upstream issue with displaying the types of properties (sphinx-doc/sphinx#7448).

In general, the generated documentation will remain sparse until more docstrings are added, however, it will hopefully still deliver an improvement over reading the sources, especially with the built-in search functionality that Sphinx provides :)

Known caveats:
- Author and derived copyright are lifted from `setup.py`; are these the most accurate to use?
- No CI integration

*<sub>Note: this is a personal contribution independent of my employer, and so I've submitted from a fork under my personal profile and email to make this distinction :heavy_check_mark:</sub>*


src/uharfbuzz/__init__.py
```diff
@@ -4,3 +4,95 @@
...
+    "ot_math_get_constant",
+    "ot_math_get_glyph_assembly",
+    "ot_math_get_glyph_italics_correction",
+    "ot_math_get_glyph_kerning",
+    "ot_math_get_glyph_kernings",
+    "ot_math_get_glyph_top_accent_attachment",
+    "ot_math_get_glyph_variants",
+    "ot_math_get_min_connector_overlap",
+    "ot_math_has_data",
+    "ot_math_is_glyph_extended_shape",
```
khaledhosny, line 89: All the `ot_*` functions above are deprecated, do we still need to include them here?

docs/source/conf.py
```diff
@@ -0,0 +1,25 @@
...
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = "uharfbuzz"
+copyright = "2024, Adrien Tétar"
+author = "Adrien Tétar"
```
khaledhosny, line 12: May be we can use something more generic, like “HarfBuzz team” or something? Adrien hasn’t been involved with this project for a while now.

khaledhosny (commented):
Looks good, thanks!. CI integration would be welcome. We probably should also publish it to readthedocs.io.

src/uharfbuzz/__init__.py
Hoolean: I've just tested locally, and if we exclude the functions here they will be excluded from documentation, but would also be excluded from any existing star-imports (i.e. `from uharfbuzz import *`), and so it would be a breaking change.

Maybe this is okay if we release explicitly as breaking and there are alternatives though? (or even as v1.0.0, if the module enjoys any production use?)

Hoolean: Alternatively there may be a way to make them show as deprecated in Sphinx

Hoolean: (in some ways, even adding `__all__` at all is already a breaking change, but the only variables being excluded at the moment are re-exports from uharfbuzz's dependencies and the standard library, which I think we can consider a bug fix)

docs/source/conf.py
Hoolean: I've now changed the `author` to `HarfBuzz Contributors` to adopt this :+1:

Hoolean: (as I am a lowly first-time contributor I've left the copyright notice as-is, but have checked "Allow edits by maintainers" on this PR in case any maintainer wishes to set it)

Hoolean:
I've added a template `.readthedocs.yml`, and have tried to use options cautiously until we are able to test in a real environment; in particular, I am not sure how the doc's dependency on a compiled wheel will be handled.

Would someone with maintainer access be willing to follow [the Read the Docs procedure](https://docs.readthedocs.io/en/stable/tutorial/index.html#creating-a-read-the-docs-account) required to create a sub-domain and have it insert actions into this repository to populate the docs?

Side note: if [the pull request option](https://docs.readthedocs.io/en/stable/tutorial/index.html#triggering-builds-from-pull-requests) is available for community plans then we may be able to avoid the duplicate runner effort of including a CI job for docs on GitHub too :rocket:

khaledhosny:
I’m merging this now to see how building on read the docs goes, and we can do any amendments/fixes in another PR.

khaledhosny:
> Side note: if [the pull request option](https://docs.readthedocs.io/en/stable/tutorial/index.html#triggering-builds-from-pull-requests) is available for community plans then we may be able to avoid the duplicate runner effort of including a CI job for docs on GitHub too

I enabled it, I don’t know if it works or not.

khaledhosny:
I did a few small tweaks and it seems the docs now build https://uharfbuzz.readthedocs.io/en/latest/, time to actually start documenting the code 😄

PR #223, opened 2024-12-17
Bind hb_face_reference_table APIs

mcratt:
face.get_table_blob(tag) or face.table_blob(tag) ?

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -635,6 +635,14 @@ cdef class Face:
             raise MemoryError()
         return Blob.from_ptr(blob)
 
+    def reference_table(self, tag: str) -> Blob:
+        cdef bytes packed = tag.encode()
+        cdef hb_tag_t hb_tag = hb_tag_from_string(<char*>packed, -1)
+        cdef hb_blob_t* blob = hb_face_reference_table(self._hb_face, hb_tag)
+        if blob is NULL:
+            raise MemoryError()
+        return Blob.from_ptr(blob)
+
     @property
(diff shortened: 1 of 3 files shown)


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -635,6 +635,14 @@ cdef class Face:
             raise MemoryError()
         return Blob.from_ptr(blob)
 
+    def get_table_blob(self, tag: str) -> Blob:
```
khaledhosny, line 638:
```suggestion
    def reference_table(self, tag: str) -> Blob:
```

mcratt: ok

tests/test_uharfbuzz.py
```diff
@@ -786,6 +786,177 @@ def test_list_names_with_user_names(self, mutatorsans):
             "MutatorMathTest-Medium_Narrow_I",
         ]
 
+    def test_reference_table(self, blankfont):
```
khaledhosny, line 789: I don’t think we need this overly detailed test. It should be enough to check for the table length, and may be the binary data. No need to try to decode it.

mcratt: ok

tests/test_uharfbuzz.py
```diff
@@ -3,7 +3,7 @@
 import sys
 import platform
 import pytest
-
+import struct
```
khaledhosny, line 6: unused import

mcratt: ok

khaledhosny:
Thanks!

PR #225, opened 2024-12-20
add blob.length property

mcratt:
I found that uharfbuzz has already bind the `hb_blob_get_length` API, but it doesn't use this API and instead returns the length using `return len(self._data)`. However, this leads to an issue: after creating a blob using `hb.Blob.from_file_path`, the length obtained by `len(blob)` or `len(blob.data)` is `0`. Only after creating `hb.Face(blob)`, can the correct length be obtained through `len(face.blob)` or `len(face.blob.data)`. This causes a significant amount of disk I/O when reading many files.

```python
def test_len(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    len_blob = len(blob)
    print(f"len(blob): {len_blob}")

    face = uharfbuzz.Face(blob)
    len_face_blob = len(face.blob)
    print(f"len(face.blob): {len_face_blob}")

def test_datalen(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    len_blob = len(blob.data)
    print(f"len(blob.data): {len_blob}")

    face = uharfbuzz.Face(blob)
    len_face_blob = len(face.blob.data)
    print(f"len(face.blob.data): {len_face_blob}")

def test_length(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    blob_length = blob.length
    print(f"blob.length: {blob_length}")

    face = uharfbuzz.Face(blob)
    face_blob_length = face.blob.length
    print(f"face.blob.length: {face_blob_length}")
```

result:
```python
test_len:
fontSize: 16998496
len(blob): 0
len(face.blob): 16998496

test_datalen:
fontSize: 16998496
len(blob.data): 0
len(face.blob.data): 16998496

test_length:
fontSize: 16998496
blob.length: 16998496
face.blob.length: 16998496
```

IO:

test_len
test_datalen
test_length

### DIFF
--- a/src/uharfbuzz/_harfbuzz.pyx
+++ b/src/uharfbuzz/_harfbuzz.pyx
@@ -383,15 +383,12 @@ cdef class Buffer:
 
 cdef class Blob:
     cdef hb_blob_t* _hb_blob
-    cdef object _data
 
     def __cinit__(self, bytes data = None):
         if data is not None:
-            self._data = data
             self._hb_blob = hb_blob_create(
-                data, len(data), HB_MEMORY_MODE_READONLY, NULL, NULL)
+                data, len(data), HB_MEMORY_MODE_DUPLICATE, NULL, NULL)
         else:
-            self._data = bytes()
             self._hb_blob = hb_blob_get_empty()
 
     @staticmethod
@@ -400,9 +397,6 @@ cdef class Blob:
 
         cdef Blob wrapper = Blob.__new__(Blob)
         wrapper._hb_blob = hb_blob
-        cdef unsigned int blob_length
-        cdef const_char* blob_data = hb_blob_get_data(hb_blob, &blob_length)
-        wrapper._data = blob_data[:blob_length]
         return wrapper
 
     @classmethod
@@ -417,17 +411,19 @@ cdef class Blob:
 
     def __dealloc__(self):
         hb_blob_destroy(self._hb_blob)
-        self._data = None
 
     def __len__(self) -> int:
-        return len(self._data)
+        return hb_blob_get_length(self._hb_blob)
 
     def __bool__(self) -> bool:
-        return bool(self._data)
+        return len(self) > 0
 
     @property
     def data(self) -> bytes:
-        return self._data
+        """Return the blob's data as bytes."""
+        cdef unsigned int blob_length
+        cdef const_char* blob_data = hb_blob_get_data(self._hb_blob, &blob_length)
+        return blob_data[:blob_length]
 
 
 class OTVarAxisFlags(IntFlag):


khaledhosny:
The whole `_data` dance seems broken to me. I don’t think we should keep it around at all. Instead we should create the blob in any of the ways currently supported, and then return data and length using the C API exclusively.

mcratt:
I'm hesitant to modify the `_data` parameter because `hb_blob_get_length` only accepts the `blob`, not `blob.data`. If I modify the `__len__()` method, it might cause errors in the new version since previous code may have used `len(blob.data`)

mcratt:
Modifying `__len__` seems to work, and `len(blob)` already produces the correct result.

```python
def __len__(self) -> int:
    return hb_blob_get_length(self._hb_blob)
```

code:
```python
def test_len(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    len_blob = len(blob)
    print(f"len(blob): {len_blob}")

    face = uharfbuzz.Face(blob)
    len_face_blob = len(face.blob)
    print(f"len(face.blob): {len_face_blob}")

def test_datalen(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    len_blob = len(blob.data)
    print(f"len(blob.data): {len_blob}")

    face = uharfbuzz.Face(blob)
    len_face_blob = len(face.blob.data)
    print(f"len(face.blob.data): {len_face_blob}")
```

result:
```python
test_len:
fontSize: 16998496
len(blob): 16998496
len(face.blob): 16998496

test_datalen:
fontSize: 16998496
len(blob.data): 0
len(face.blob.data): 16998496
```

mcratt:
If `self._data` is not used for storage, perhaps `hb_blob_get_data` can be used to retrieve the data, but I'm not sure if this will cause other issues.

mcratt:
It seems to be running well, although reading the data part inevitably involves accessing the hard drive, as the data is in bytes.

```python
def test_len(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    print(f"len(blob): {len(blob)}")

    face = uharfbuzz.Face(blob)
    print(f"len(face.blob): {len(face.blob)}")

def test_datalen(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    blob = uharfbuzz.Blob.from_file_path(fontPath)
    print(f"len(blob.data): {len(blob.data)}")

    face = uharfbuzz.Face(blob)
    print(f"len(face.blob.data): {len(face.blob.data)}")

def test_face_blob(fontPath):
    fontSize = os.path.getsize(fontPath)
    print(f"fontSize: {fontSize}")

    with open(fontPath, "rb") as f:
        fontBytes = f.read()
    face = uharfbuzz.Face(fontBytes)
    print(f"len(face.blob): {len(face.blob)}")
    print(f"len(face.blob.data): {len(face.blob.data)}")
```

result:
```python
test_len:
fontSize: 16998496
len(blob): 16998496
len(face.blob): 16998496

test_datalen:
fontSize: 16998496
len(blob.data): 16998496
len(face.blob.data): 16998496

test_face_blob:
fontSize: 16998496
len(face.blob): 16998496
len(face.blob.data): 16998496
```

khaledhosny:
> although reading the data part inevitably involves accessing the hard drive

If the blob is created from a file path, HarfBuzz will mmap the file, so reading the data will indeed involve file I/O.

mcratt:
Yes, but now the build is failing, and it's strange because I can't find where the issue is. Is it possible that `_hb_blob` is not being cleaned up properly?

src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -383,15 +383,12 @@ cdef class Buffer:
 
 cdef class Blob:
     cdef hb_blob_t* _hb_blob
-    cdef object _data
 
     def __cinit__(self, bytes data = None):
         if data is not None:
-            self._data = data
             self._hb_blob = hb_blob_create(
                 data, len(data), HB_MEMORY_MODE_READONLY, NULL, NULL)
```
khaledhosny, line 390:
```suggestion
                data, len(data), HB_MEMORY_MODE_DUPLICATE, NULL, NULL)
```

Since we are no longer keeping `data` around, so we need to ask HarfBuzz to make a copy.

khaledhosny:
Thanks!

mcratt:
Looking forward to the new version.

Issue #227, opened 2025-01-01
Path toward not requiring experimental harfbuzz APIs?


musicinmybrain:
I’m working on packaging `uharfbuzz` for Fedora so that we can offer the `repacker` extras in our `fonttools` and `fontmake` packages.

The support for using a system copy of `harfbuzz`, added in https://github.com/harfbuzz/uharfbuzz/pull/201, is extremely helpful here. However, I’ve had to ask the maintainers of Fedora’s `harfbuzz` package to [enable experimental APIs](https://src.fedoraproject.org/rpms/harfbuzz/pull-request/9). It looks like that might be possible, but there seems to be an understandable desire not to do this indefinitely.

(Using the bundled `harfbuzz` sources [is also an option](https://docs.fedoraproject.org/en-US/packaging-guidelines/#bundling) under Fedora policies, if necessary, but using the system copy of a library is always preferable.)

If https://github.com/harfbuzz/uharfbuzz/issues/45#issuecomment-2267947172 still tells the whole story, `uharfbuzz` needs experimental APIs for `hb_subset_repack_or_fail()`. Is there any path toward stabilization of this API in a future `harfbuzz` release, or any other path toward eventually being able to link `uharfbuzz` against a `harfbuzz` library that was built without experimental APIs?

behdad:
I think we can remove the repacking API from experimental to stable.

musicinmybrain:
Thanks for investigating! I will put `python-uharfbuzz` packaging on the back burner for a little while and wait to see what happens here.

garretrieger:
> I think we can remove the repacking API from experimental to stable.

Sounds good to me.

garretrieger:
#4987 and #4989 promoted the repacker api from experimental to stable. Note we also changed the naming so uharfbuzz will need to be updated when it picks up the next hb release.

musicinmybrain:
Thanks to stablization of the repacker API in https://github.com/harfbuzz/harfbuzz/releases/tag/10.2.0, and to the corresponding adjustments in https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.45.0, I just submitted a `python-uharfbuzz` package for review in Fedora. Thanks again for working on this!

Issue #203, opened 2024-08-23
Add a CI job to test building with unvendored harfbuzz USE_SYSTEM_LIBS=1


anthrotype:
https://github.com/harfbuzz/uharfbuzz/pull/201 added support for optionally linking uharfbuzz with a preinstalled harfbuzz library, this works but is not currently tested on the CI.
To ensure this continues to work we should add a job to test building with USE_SYSTEM_LIBS=1, at least on Linux where this is most likely to be used.

PR #228, opened 2025-01-11
Update to HarfBuzz 10.2.0

khaledhosny:


justvanrossum:
FontGoggles was proudly supporting the experimental VARC table, but this change silently broke that. Any chance we can get experimental API support back?

khaledhosny:
Not using experimental HarfBuzz APIs was requested in #227. But I think the actual concern there was actually using experimental APIs, which break when linking with system HarfBuzz, but since VARC support does not expose any APIs, I think we can put the `HB_EXPERIMENTAL_API` back when building with vendored HarfBuzz.

khaledhosny:
We will have to be declined and not use any experimental APIs, though, as this will not be caught by CI until we have #203 addressed.

khaledhosny:
https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.48.0

PR #241, opened 2025-05-08
Update ci.yml for aarch64 linux build

### LINKED ISSUE #240: Add pre-built wheels for linux aarch64
I'm running python 3.13 on linux under docker on ARM.

The lack of a linux aarch64 wheel increases my build time by over 4 minutes!

This is the only major dependency that doesn't supply these wheels.

Can this be added to the build pipeline please?

HinTak:
Adding aarch64 linux build to the matrix

Hopefully this is sufficient to fix #240

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


HinTak:
It ie failing with dependency on ccache, which I think should be optional. Anyway, I 'll let others to figure that out.

Issue #254, opened 2025-09-03
Use PyPI action to publish


EpicWink:
Switch to [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) for the "Publish to PyPI" job in the CI workflow, instead of direct invocation of `twine upload`, using trusted publishing.

This adds [build artefact attestations](https://docs.pypi.org/attestations/) in the upload to PyPI. Trusted publishing (with attestations) means I can know for certain that what I download from PyPI is the same artefact which was generated in GitHub CI, meaning that what I see in GitHub is the same as what is installed - handy for auditing (rather than having to manually review all of the installed files on each release).

See [the Python packaging documentation](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#configuring-trusted-publishing), [the PyPI documentation](https://docs.pypi.org/trusted-publishers/), and [the official pypi-publish GitHub action documentation](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) on trusted publishing - you'll need to configure an environment in PyPI and GitHub. You will be able to remove the `PYPI_USERNAME` and `PYPI_USERNAME` project secrets.

PR #256, opened 2025-09-22
Use trusted publishing

madig:
In light of the recent npm supply chain attacks and also https://blog.pypi.org/posts/2025-09-16-github-actions-token-exfiltration/, I'm combing through our font stack to see if all them Py projects are using the trusted publisher mechanism as recommended by PyPI. See https://docs.pypi.org/trusted-publishers/ and https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi.

Someone needs to do two things for this PR to work:

* Create an environment called "publish-to-pypi" in this GitHub repository under Settings -> Environments. Creating alone is probably enough, no configuration needed I think.
* Follow https://docs.pypi.org/trusted-publishers/adding-a-publisher/ to set up the other side on PyPI.

I'm not sure if one needs to do anything to make twine pick up the new creds, trusted publishing should be supported in v6.1.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,12 @@ jobs:
     if: contains(github.ref, 'refs/tags/')
     env:
       GH_TOKEN: ${{ github.token }}
+    environment:
+      name: publish-to-pypi
+      url: https://pypi.org/p/uharfbuzz
+    permissions:
+      id-token: write # IMPORTANT: mandatory for trusted publishing
+
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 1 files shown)


khaledhosny:
Does this PR address #254 as well? I can merge this, but I don’t have access to the PyPI project to finish the required steps there.

anthrotype:
>  I don’t have access to the PyPI project to finish the required steps there.

I can do those. Hold on

anthrotype:
ok, I just did the PyPI setting. Are we sure twine supports trusted publishing? I suppose it must do. Otherwise I'd simply use the official PyPA publish github action like I did in https://github.com/fonttools/openstep-plist/commit/166bab31c6b4ea2e69090b865c6fd6dec07522dd

khaledhosny:
> Otherwise I'd simply use the official PyPA publish github action like I did in [fonttools/openstep-plist@166bab3](https://github.com/fonttools/openstep-plist/commit/166bab31c6b4ea2e69090b865c6fd6dec07522dd)

I think this is what #254 requests. I don’t mind if we do it in this PR as well.

anthrotype:
my test pre-release with trusted publishing using pypa GH action worked, see https://pypi.org/project/openstep-plist/0.5.1b2/#files

I suggest we do the same here, build the sdist (e.g. `pipx run build --sdist`) and then use the official pypa/gh-action-pypi-publish

PR #258, opened 2025-10-02
Bind hb_subset_input_override_name_table API as SubsetInput.override_…

mcratt:
This API is marked as `HB_EXPERIMENTAL_API`.


src/uharfbuzz/_harfbuzz.pyx
```diff
@@ -2955,6 +2956,20 @@ cdef class SubsetInput:
...
+        if name_str is None:
+            c_name = NULL
+            str_len = -1
+        else:
+            c_name = PyUnicode_AsUTF8AndSize(name_str, &str_len)
+            if c_name == NULL:
+                raise MemoryError()
+        return hb_subset_input_override_name_table(
+            self._hb_input, name_id, platform_id, encoding_id, language_id, c_name, str_len
+        )
```
khaledhosny, line 2971: Linux distributions builds want to link to system HarfBuzz which is typically built without experimental APIs, so this call should be wrapped in `HB_EXPERIMENTAL_API` and raise a `NotImplementedError` otherwise.

khaledhosny:
I’m a bit reluctant to expose this. See also https://github.com/harfbuzz/harfbuzz/issues/5595.

behdad:
If we address https://github.com/harfbuzz/harfbuzz/issues/5595 first, we can graduate the API out of experimental.

PR #264, opened 2025-10-17
enable Py_LIMITED_API to strip down build matrix to one abi3 wheel per platform

### LINKED ISSUE #262: Investigate use of Py_LIMITED_API for cython modules to use one wheel per platform
Prompted by #260 

For reference https://cython.readthedocs.io/en/latest/src/userguide/limited_api.html

anthrotype:
- Enabled Python Limited API for stable ABI (abi3) support. We now build a single cp310-abi3 wheel for each platform which works across Python 3.10 and above. This significantly reduces number of wheels we distribute on PyPI.
- For macOS CPython builds, we now publish a single 'fat' universal2 wheel (Intel + Apple Silicon), that's sufficient, we don't need separate wheels for arm64 and x86_64.
- For PyPy, we still need to build separate x86_64 and arm64 wheels, and also disable the Limited API because that's CPython-only. But we only distribute one wheel for pypy3.11 so that's fine.
- Dropped Python 3.8 and 3.9 support (minimum now Python 3.10+)

Fixes #262

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,8 @@ build-backend = "setuptools.build_meta"
 [tool.cibuildwheel]
 # Also skip wheels for free-threaded CPython 3.14 (e.g. 'cp314t') until we
 # actually support them. We currently aren't thread-safe as we define HB_NO_MT=1.
-skip = ["cp38-*", "cp39-*", "cp3??t-*"]
+# CPython abi3 wheels only need universal2 on macOS, skip the single-arch builds.
+skip = ["cp38-*", "cp39-*", "cp3??t-*", "cp*-macosx_x86_64", "cp*-macosx_arm64"]
 enable = ["pypy"]
 test-requires = "pytest"
 test-command = "pytest {project}/tests"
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -32,6 +32,12 @@ def bool_from_environ(key: str):
 use_system_libraries = bool_from_environ("USE_SYSTEM_LIBS")
 use_cython_linetrace = bool_from_environ("CYTHON_LINETRACE")
 use_cython_annotate = bool_from_environ("CYTHON_ANNOTATE")
+# Python Limited API for stable ABI support is enabled by default.
+# Set USE_PY_LIMITED_API=0 to turn it off.
+# https://docs.python.org/3.14/c-api/stable.html#limited-c-api
+use_py_limited_api = bool_from_environ("USE_PY_LIMITED_API", default=True)
+# NOTE: this must be kept in sync with python_requires='>=3.8' below
+limited_api_min_version = "0x03080000"
```
madig: I think you can require >= 3.10, because 3.9 is going to be EOL this month?

anthrotype: probably. I kept the current python_requires to minimize changes

PR #275, opened 2026-01-02
Pyodide support

### LINKED ISSUE #270: Pyodide compatibility?
Hey,

I am working on enabling [some projects](https://github.com/pygfx/wgpu-py/pull/753) to run in the browser via pyodide. I wanted to ask if you would consider building `uharfbuzz` for wasm so it can run in pyodide as well as listing it in the pyodide distribution? The goal being existing python projects to run without any browser specific changes. 

I have found [one mention](https://github.com/googlefonts/fontquant/issues/17#issuecomment-2836888904) of this already working in the past.

anthrotype:
- Detect Emscripten cross-compilation in setup.py, skipping platform-specific sources (CoreText, DirectWrite, Uniscribe)
- Add build-pyodide job to CI workflow
- Upload Pyodide wheel to GitHub releases only (PyPI doesn't accept wasm32 wheels yet, see https://github.com/pypi/warehouse/issues/10416 and https://discuss.python.org/t/pep-783-emscripten-packaging)
- Also add `tox -e pyodide` for local build and test

Fixes #270

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,8 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: "3.x"
+        # Pyodide requires Python 3.13 (latest not yet supported)
+        python-version: "3.13"
     - name: Install cibuildwheel
       run: pip install cibuildwheel
     - name: Build Pyodide wheel
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,9 +36,35 @@ jobs:
         name: uharfbuzz-${{ matrix.os }}
         path: wheelhouse/*.whl
 
+  build-pyodide:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.x"
+    - name: Install cibuildwheel
+      run: pip install cibuildwheel
+    - name: Build Pyodide wheel
+      run: python -m cibuildwheel --platform pyodide --output-dir wheelhouse
+    - name: Set up Pyodide venv and run tests
+      run: |
+        pip install pyodide-build
+        pyodide venv .pyodide-venv
+        .pyodide-venv/bin/pip install pytest
+        .pyodide-venv/bin/pip install wheelhouse/*.whl
+        .pyodide-venv/bin/python -m pytest tests/ -v
+    - uses: actions/upload-artifact@v4
+      with:
+        name: uharfbuzz-pyodide
+        path: wheelhouse/*.whl
+
   deploy:
     name: Create and populate release
-    needs: build
+    needs: [build, build-pyodide]
     runs-on: ubuntu-latest
     if: contains(github.ref, 'refs/tags/')
     env:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -96,50 +105,57 @@ def _configure_extensions_with_vendored_libs() -> List[Extension]:
...
 
-    extra_compile_args = []
-    extra_link_args = []
-    libraries = []
-    if platform.system() != "Windows":
+    if platform.system() == "Windows" and not emscripten:
+        define_macros.append(("HAVE_DIRECTWRITE", "1"))
+        define_macros.append(("HAVE_UNISCRIBE", "1"))
+        libraries += ["usp10", "gdi32", "user32", "rpcrt4", "dwrite"]
+        sources += [
```
behdad: I think `harfbuzz.cc` has all these files automatically included.

behdad: Right. That's annoying.

khaledhosny:
Uploading the Pyodide wheels seems to fail: https://github.com/harfbuzz/uharfbuzz/actions/runs/23300214344/job/67760947291

```
Run if [ -d dist-pyodide ] && [ "$(ls -A dist-pyodide)" ]; then
  if [ -d dist-pyodide ] && [ "$(ls -A dist-pyodide)" ]; then
    gh release upload refs/tags/v0.53.4 dist-pyodide/*.whl
  fi
  shell: /usr/bin/bash -e {0}
  env:
    GH_TOKEN: ***
    pythonLocation: /opt/hostedtoolcache/Python/3.14.3/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.14.3/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.14.3/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.14.3/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.14.3/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.14.3/x64/lib
    RELEASE_NAME: 0.53.4
    IS_PRERELEASE: false
release not found
```

anthrotype:
You can't upload to PyPI yet I think, they aren't sufficiently standardised IIRC. I am at a conference right now and can't check, but I believe the path forward for now would be to disable the automatic upload for those and perhaps make them available only as GitHub Release downloadable assets

anthrotype:
Oh now I remember I had written about that (see above here)

> Upload Pyodide wheel to GitHub releases only (PyPI doesn't accept wasm32 wheels yet, see Add support for wasm32 platform pypi/warehouse#10416 and https://discuss.python.org/t/pep-783-emscripten-packaging)

anthrotype:
Oh! It's the GH assets upload that it's failing (not PyPI), sorry I hadn't read it carefully. Ok not sure. Next thing to check, is the GitHub token used enabled for "write"?

.github/workflows/ci.yml
```diff
@@ -106,6 +131,11 @@ jobs:
                             --title "${{ env.RELEASE_NAME }}" \
                             $notes \
                             $prerelease
+    - name: Upload Pyodide wheel to GitHub release
+      run: |
+        if [ -d dist-pyodide ] && [ "$(ls -A dist-pyodide)" ]; then
+          gh release upload ${{ github.ref }} dist-pyodide/*.whl
```
anthrotype, line 137: @khaledhosny i think the issue is this is supposed to be github.ref_name ("v0.53.4") instead of github.ref ("refs/tags/v0.53.4").
And for consistency also the gh release create call right above this (which somehow normalises ref to ref_name internally, whereas upload command doesn't).

anthrotype: since we'll have to re-tag, the release that was created will have to be deleted as well

khaledhosny: Thanks, I fixed it now.

Vipitis:
> I was waiting for them to test if these actually work.. the tests pass, so feel free to merge

sorry for missing this. Best I can tell harfbuzz and shaping works on pyodide. (I didn't get to implement glyph rendering, but that isn't using harfbuzz yet).

PR #289, opened 2026-04-27
Switch build from harfbuzz-subset.cc to harfbuzz-world.cc

khaledhosny:
The harfbuzz-world.cc is HarfBuzz new amalgam build file, and it allows building more HarfBuzz libraries from a single source file.

PR #290, opened 2026-04-27
Bind harfbuzz-raster APIs

khaledhosny:
Does not include PNG support, though.


khaledhosny:
Should be merged after #289

behdad (approved):

PR #292, opened 2026-04-28
When using system Harfbuzz, link the harfbuzz-raster component

musicinmybrain:
This fixes errors like `ImportError: […]/uharfbuzz/_harfbuzz.abi3.so: undefined symbol: hb_raster_image_get_extents
` since 29d1e9634ee383739a7d5fb992534d5457927f0d.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ def _configure_extensions_with_system_libs() -> List[Extension]:
     libraries = []
     library_dirs = []
 
-    harfbuzz_components = ["harfbuzz-subset"]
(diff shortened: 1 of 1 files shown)


khaledhosny:
Thanks!

alerque:
This probably merits a quick patch release; the current release as published in various packages is basically unusable without hackery.

khaledhosny:
https://github.com/harfbuzz/uharfbuzz/releases/tag/v0.54.1 (artifacts and github/pypi release should happen in a bit)

PR #314, opened 2026-09-05
Add type stub generation to Harfbuzz

vibbix:
# Background
While working around with `fonttools` and such, the lack of proper type annotations for `harfbuzz` started driving me a bit crazy. I tried running it manually in my repos, but I noticed it had type issues upstream, so I wrote this PR.

# Changes
 - Emulated Nvidia-Cuda's `stubgen-pyx` usage, in `setup.py`
 - dropped `Set[int]` to just `Set` - checked by `mypy`
   - I attempted to get it to extend `MutableSet[Int]` but that was quite troublesome
- fixed an errant `;`
- `Set` now returns `SetIter` properly
- Removed an errant forward declaration of `hb_color_line_t` in `src/uharfbuzz/charfbuzz.pxd` 
- added `uv.lock` to ignores, alongside `pyi` and `.typed`
- misc. small tweaks to improve `mypy` detection

# Issues
1. *major* Currently, `stubgen-pyx` is a hit behind, so the use of `cython 3.3.0` isn't quite here yet. 

# LLM Disclosure
All work was done by me, but I had Gemini/AGY double check things for me.


vibbix:
If it's preferred, I can switch this to a `github stack` that splits the fixes to `pyx` types from the `binding` generator addition, so they can be reviewed separately.

khaledhosny:
As a start, give commit messages more descriptive subjects and bodies so one knows what each commit is actually about.

